PR #65, opened 2015-01-15
Add support for logging request & response headers (in logstash-access)

vbehar:
Disabled by default (because it can be quite big), but it can be activated with:

      <fieldNames>
        <fieldsRequestHeaders>@fields.request_headers</fieldsRequestHeaders>
        <fieldsResponseHeaders>@fields.response_headers</fieldsResponseHeaders>
      </fieldNames>
    </encoder>

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,5 +400,16 @@ The default field names used for access logs are different than those documented
 See [`LogstashAccessFieldNames`](/src/main/java/net/logstash/logback/fieldnames/LogstashAccessFieldNames.java)
 for all the field names used for access logs.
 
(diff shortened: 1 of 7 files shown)


philsttr:
Hi @vbehar,

Is it normally the case where you want _all_ request/response headers logged?

It seems like you'd want to filter these somehow.

vbehar:
Hi,

yes, in our case, we want to log ALL request/request headers (for hadoop-like processing/analysis).

but I can understand that some people will be interested in only a few headers. Do you want me to add a filtering feature ?

philsttr:
Ok.  No, if you don't have an immediate need, we can delay that for now.  I'll merge your pull request into master

PR #67, opened 2015-01-18
Added mdc fields white list filtering

sha1n:
Hi,

I created this pull request since I need to be able to control which MDC properties are logged and which are not. The reason is that I live in a framework which uses MDC heavily and I want to create a very specific and lightweight log, so I need to filter out most of the framework's MDC information.

Your feedback is appreciated.

Thanks,
Shai

### DIFF
--- a/src/main/java/net/logstash/logback/LogstashFormatter.java
+++ b/src/main/java/net/logstash/logback/LogstashFormatter.java
@@ -13,30 +13,25 @@
  */
 package net.logstash.logback;
 
-import java.io.IOException;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Map.Entry;
-
-import net.logstash.logback.fieldnames.LogstashFieldNames;
-import net.logstash.logback.marker.LogstashMarker;
-import net.logstash.logback.marker.Markers;
-
-import org.slf4j.MDC;
-import org.slf4j.Marker;
-
 import ch.qos.logback.classic.pattern.Abbreviator;
 import ch.qos.logback.classic.pattern.ExtendedThrowableProxyConverter;
 import ch.qos.logback.classic.pattern.TargetLengthBasedClassNameAbbreviator;
 import ch.qos.logback.classic.pattern.ThrowableHandlingConverter;
-import ch.qos.logback.classic.pattern.ThrowableProxyConverter;
 import ch.qos.logback.classic.spi.ILoggingEvent;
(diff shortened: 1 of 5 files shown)


philsttr:
Hi @sha1n,

I love the idea, but I have an alternate implementation proposal.

I didn't like how the list was configured as a comma separated string.  I'd prefer to use the logback's standard way to configure lists (e.g. one xml element per list element).

I didn't like how the list was configured on the `LogstashFieldNames`, since the primary purpose of that class is for configuring standard field names.

So, I've created an alternative implementation available on the [mdc_entry_filtering](https://github.com/logstash/logstash-logback-encoder/tree/mdc_entry_filtering) branch. It does the following:

* Uses logback's standard way of configuring lists (e.g. one xml element per list element)
* Allows inclusion or exclusion of specific fields
* Configures the inclusion/exclusion rules on the encoder, rather than the `LogstashFieldNames`

Please take a look at the branch, and [the way its configured](https://github.com/logstash/logstash-logback-encoder/tree/mdc_entry_filtering#mdc-fields), and let me know if this will satisfy your requirements.

Thanks.

sha1n:
Hi @philsttr,

Thank you very much for the quick response! I reviewed your changes and tested them and it seems all good.
A few comments:
1. I didn't like the way the list was configured myself, but I didn't know you could work with lists that way in logback's config - so I'm glad I learned that for next time :)
2. Regarding `LogstashFieldNames`, well my first implementation was in the formatter as well, but then I thought about it and figured that a formatter is responsible for formatting, so theoretically, if I use another formatter, my log will still contain the same bits of data. I realize this is not very relevant here, since this formatter is a logstash thing and it's unlikely that you will use another formatter, but that's just what I had in mind.

So, as far as I'm concerned, you can merge your changes into master. If there is anything I can do to help, just let me know.

Thanks.

philsttr:
Thanks @sha1n,

Yeah, for 2, I agree that it's not designed as well at it could be.  I've been thinking about pulling some of the logic out of the formatters and using a composition design, similar to how PatternLayout doesn't actually do any formatting itself, it delegates to composed converters.  I think if the LogstashFormatter followed this design, it would be cleaner to implement this issue.  i.e. all of the MDC related things would be in something that is only responsible for the MDC, and nothing else.

However, without having that design in place, I think the best place for this logic is in the formatter itself.

I'll merge to master.

PR #68, opened 2015-01-20
Add configurable layouts/encoders

dimas:
This PR adds a new layouts (+wrapping encoders) for both ILoggingEvent and IAccessEvent.
The layout pattern allows providing your own JSON where each text value will be run through logback's standard PatternLayout so all things like %date, %thread etc will be substituted.

It is all in the README but the idea is that you can do something like:

```xml
<encoder class="net.logstash.logback.encoder.LoggingEventJsonPatternLayoutEncoder">
    <pattern>
        {
        "custom_constant": "123",
        "tags": ["one", "two"],

        "logger": "%logger",
        "level": "%level",
        "thread": "%thread",
        "message": "%m",
...
        }
    </pattern>
</encoder>
```

The main reason it was created is for IAccessEvent actually because it allows picking individual cookies, request headers and attributes.


philsttr:
Hi Dimetry.

Awesome idea, and I like this a lot, especially since you can take advantage of everything available in `PatternLayout`.

I'm in the middle of refactoring the existing formatters/encoders/layouts to make them more composable.  I.e. they will become just containers for a bunch of other providers that provide various fields into the JSON output.

I think this will "plug in" to this new composable formatter/encoder/layout very easily.  I'd like to wait until the new composable stuff is in place before incorporating your pull request.  The composable stuff should be available sometime this week.

dimas:
I merged in latest master from you and fixed the build which I broke.

I am not in a rush - for now we will be using a custom build from my fork.
Although it looks to me safe to merge and adapt encoders/layouts later when they are available.
While fixing  conflicts in README I read the new  changes there that happened since I forked - the timezone support and providing the list of headers to record. Both of these can be handled by the new encoder without any changes "%date{ISO8601, timezone}" and "%reqHeader{name}" respectively.

To be honest, the only thing I needed for myself was request attributes so my first impulse was to do very much what was done for headers - adding a new configuration option to the existing encoder. But then I though it is a bad idea to keep overloading that encoder with more option which will solve my immediate problem but won't add flexibility. Unfortunately that thought took me for a long way...

Re formatters/layouts - initially I wanted to implement alternative formatter and allow changing it for the encoder. But configuration becomes a bit heavy so in the end I ditched formatter and did the same thing logback does - there is a layout and encoder wrapping it.

philsttr:
I've committed the initial composite encoder work to the [composition](https://github.com/logstash/logstash-logback-encoder/tree/composition) branch.

I incorporated your changes under a new `pattern` JSON provider.  It can be included in the composite encoders/layouts along with any/all of the other providers.

This unifies all the encoders/layouts available in logstash-logback-encoder.  LogstashEncoder/Layout and LogstashAccessEncoder/Layout are effectively just pre-defined compositions now.

A couple things to note..
* I removed the `mdc` operation, since you can just include the `mdc` provider now.
* And I removed the behaviour of including the timestamp/version by default (since the pattern provider does not control the whole event now; it only contributes to it.

I have updated the readme on the branch with all the new details.

I'm still tieing up a few loose ends, but the majority of the work is complete.  Hope to have this finalized in the next few days.

dimas:
Hi.
Couple of things:
1. AbstractJsonPatternParserTest.shouldAllowOverridingDefaultFields() - you probably want to remove this test because it has lost its purpose
2. I haven't looked thoroughly but my first impression was that pattern provider will only be able to generate a nested object because it returns ObjectWriter which, when its write(0 is invoked, will produce the complete object JSON including opening and closing curly brace.

philsttr:
Excellent eye!

I've corrected both.

I'm going to do some more testing, and write some more unit tests for the composition stuff.

PR #96, opened 2015-05-21
Update jackson version to 2.5.0

adrianchung:


philsttr:
I can certainly update jackson to the latest version (2.5.1) in logstash-logback-encoder.

However, if you're only updating it here so that your project uses the later jackson version, I'd suggest just declaring the dependency on jackson in your project's pom (or using `dependencyManagement` in your project's pom) if you want to control exactly which jackson version is used in your project.

adrianchung:
Yup that's cool. I was running into IntelliJ classpath issues picking up the wrong version of jackson I wanted used and figured it's be easiest to just push the ball forward getting the version updated. I can work around it in the meantime.

philsttr:
Updated to 2.5.4

PR #102, opened 2015-06-14
fixed setting a prefix without a suffix on the CompositeJsonLayout

vboulaye:
parentheses were missing in the way the length of the message was computed when setting only a prefix or a suffix layout inside a CompositeJsonLayout.


philsttr:
Thanks!  Incorporated for 4.4

PR #108, opened 2015-06-22
add BasicMarker logic to LogstashBasicMarker

melrief:
Also, because LogstashBasicMarker doesn't depend from BasicMarker anymore, move LogstashBasicaMarker to logstash package


philsttr:
Merged.  Thanks!

BTW, I added licensing information to `LogstashBasicMarker` in a followup commit.

Issue #115, opened 2015-07-28
TCPSocketAppender with failover


brenuart:
I'm looking for a way to add some failover behaviour to the `LogstashTCPSocketAppender` such that it would temporarily connect to a secondary server when the primary is not reachable. When connected to the secondary, the appender would retry on the master at regular interval until it succeeds.

We are running multiple services per host in a container environment. A log collector is running on every host - services are configured to send their log to the local log collector. If not available, then they should connect to another whose only purpose is to serve as fallback when one of the primaries fail. This setup allows us to shutdown temporarily the local log collector for maintenance without affecting the log collection.

As far as I can't tell such feature is not supported out-of-the box by the `LogstashTCPSocketAppender`. We had a look at the current implementation to see if we could somehow customise it to match our needs but unfortunately most of the relevant classes are either private or package protected. 

We were wondering if such feature is of interest to other people and if you have plans to implement it soon or later.

philsttr:
Hi @brenuart ,

Have you considered using a VIP/load balancer?  I ask, because I have most often seen this use case (not just for logging, but any TCP service) handled on the destination side, without complex logic being required of clients.

brenuart:
Yes of course but we are in a situation where customer don't want to hear about such solution :-(
Stupid, 'cause a simple HAProxy could do the job. So we currently have no other choice than a software solution inside the logging client.

philsttr:
Ok.  Understood.

I'm somewhat hesitant to add this functionality to logstash-logback-encoder, since it is not the best practice to do this within a client side library.  

However, please feel free to submit a PR, and I'll take a look.  If it's not terribly complex, I'll merge it.

I'm thinking it could be as simple as changing `AbstractLogstashTcpSocketAppender.TcpSendingEventHandler.openSocket()` to go down a list of `remoteHost` names as it loops, rather than a single remoteHost.  It would connect to the first one that accepts the connection.

Then, you can control the behavior on the server side by:
* bring up secondary host
* take down primary host
* appender will reconnect to secondary
* perform maintenance on primary
* bring up primary host
* take down secondary host
* appender will reconnect to primary

brenuart:
Since `TcpSendingEventHandler` is a private inner class, I was currently playing at the `SocketFactory` level. It is a big ugly I admit, but that's the only solution I found so far.

[...]

PR #116, opened 2015-07-28
TcpSocketAppender Failover implementation

brenuart:
See https://github.com/logstash/logstash-logback-encoder/issues/115

This PR holds a first quick-and-dirty attempts to implement the feature. Please, have a look for review and comments.

##### Short description of changes
A comma separated list of `host:port` pairs can be specified via the `<hostAddresses>` configuration property. The first host is the primary, while the others are considered secondaries.

The appender tries to connect to the primary and fallbacks to a secondary when unavailable. Hosts are tried in the order they appear in the `<hostAddresses>` property.

When connected to a secondary, the appender tries to reconnect to the primary after `<reconnectionToPrimaryDelay>` time.

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
@@ -32,6 +32,8 @@ import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 
+import org.apache.commons.lang.StringUtils;
+
 import net.logstash.logback.encoder.SeparatorParser;
 import ch.qos.logback.core.encoder.Encoder;
 import ch.qos.logback.core.net.ssl.ConfigurableSSLSocketFactory;
@@ -80,6 +82,11 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
     public static final int DEFAULT_RECONNECTION_DELAY = 30000;
 
     /**
+     * The default reconnection to primary server delay (60000 milliseconds or 60 seconds)
+     */
+    public static final int DEFAULT_RECONNECTION_PRIMARY_DELAY = 60000;
+    
+    /**
      * Default size of the queue used to hold logging events that are destined
      * for the remote peer.
      * Assuming an average log entry to take 1k, this would result in the application
@@ -106,12 +113,30 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
     private int port = DEFAULT_PORT;
 
     /**
+     * Comma separated list of host addresses
+     */
+    private String hostAddresses;
+    
+    /**
+     * Array with the host/port information sorted by preference
+     */
+    private HostInfo[] hostInfo;
+    
+    /**
      * Time period for which to wait after a connection fails,
      * before attempting to reconnect.
      * Default is {@value #DEFAULT_RECONNECTION_DELAY} milliseconds.
      */
     private Duration reconnectionDelay = new Duration(DEFAULT_RECONNECTION_DELAY);
 
+    /**
+     * Time period to wait before attempting to reconnect to primary server
+     * when multiple servers are specified. Only happens when the appender
+     * is currently connected to a secondary.
+     * Default is {@value #DEFAULT_RECONNECTION_PRIMARY_DELAY} milliseconds.
+     */
+    private Duration reconnectionToPrimaryDelay = new Duration(DEFAULT_RECONNECTION_PRIMARY_DELAY);
+    
     /**
      * Socket connection timeout in milliseconds. 
      */
@@ -229,6 +254,12 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
         private volatile long lastSentTimestamp;
         
         /**
+         * Time at which the current connection should be automatically closed
+         * to force an attempt to reconnect to the primary server
+         */
+        private volatile long connectionTTL = Long.MAX_VALUE;
+        
+        /**
          * Future for the currently scheduled {@link #keepAliveRunnable}.
          */
         private ScheduledFuture<?> keepAliveFuture;
@@ -307,6 +338,15 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                         outputStream.flush();
                     }
                     lastSentTimestamp = currentTime;
+                    
+                    /*
+                     * Should we close the current connection?
+                     */
+                    if( connectionTTL <= currentTime ) {
+                    	addInfo(peerId + "closing connection and attempt to reconnect to primary server.");
+                    	outputStream.flush();
+                    	reopenSocket();
+                    }
                     break;
                 } catch (Exception e) {
                     addWarn(peerId + "unable to send event: " + e.getMessage(), e);
@@ -353,20 +393,26 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
          * then it should be able to be used to send.
          */
         private synchronized void openSocket() {
+        	int hostIndex = 0; 
             int errorCount = 0;
             while (isStarted() && !Thread.currentThread().isInterrupted()) {
                 long startTime = System.currentTimeMillis();
                 Socket tempSocket = null;
                 OutputStream tempOutputStream = null;
                 try {
-                    tempSocket = socketFactory.createSocket();
+                	/*
+                	 * Choose next server and update peerId (for status message)
+                	 */
+                    peerId = "Log destination " + hostInfo[hostIndex] + ": ";
+                    
                     /*
                      * Set the SO_TIMEOUT so that SSL handshakes will timeout if they take too long.
                      * 
                      * Note that SO_TIMEOUT only applies to reads (which occur during the handshake process).
                      */
+                    tempSocket = socketFactory.createSocket();
                     tempSocket.setSoTimeout(acceptConnectionTimeout);
-                    tempSocket.connect(new InetSocketAddress(remoteHost, port), acceptConnectionTimeout);
+                    tempSocket.connect(new InetSocketAddress(hostInfo[hostIndex].host, hostInfo[hostIndex].port), acceptConnectionTimeout);
                     tempOutputStream = new BufferedOutputStream(tempSocket.getOutputStream(), writeBufferSize);
                     
                     encoder.init(tempOutputStream);
@@ -375,6 +421,18 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                     
                     this.socket = tempSocket;
                     this.outputStream = tempOutputStream;
+                    
+                    /*
+                     * If connected to a secondary, remember when the connection should be closed to
+                     * force attempt to reconnect to primary
+                     */
+                    if( hostIndex != 0 ) {
+                    	connectionTTL = startTime + reconnectionToPrimaryDelay.getMilliseconds();
+                    }
+                    else {
+                    	connectionTTL = Long.MAX_VALUE;
+                    }
+                    
                     return;
                     
                 } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -80,6 +82,11 @@
     public static final int DEFAULT_RECONNECTION_DELAY = 30000;
 
     /**
+     * The default reconnection to primary server delay (60000 milliseconds or 60 seconds)
+     */
+    public static final int DEFAULT_RECONNECTION_PRIMARY_DELAY = 60000;
```
philsttr: Hmm.  I think the default should be to not attempt to re-establish the connection.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -113,6 +130,14 @@
     private Duration reconnectionDelay = new Duration(DEFAULT_RECONNECTION_DELAY);
 
     /**
+     * Time period to wait before attempting to reconnect to primary server
+     * when multiple servers are specified. Only happens when the appender
+     * is currently connected to a secondary.
+     * Default is {@value #DEFAULT_RECONNECTION_PRIMARY_DELAY} milliseconds.
+     */
+    private Duration reconnectionToPrimaryDelay = new Duration(DEFAULT_RECONNECTION_PRIMARY_DELAY);
```
philsttr: The naming of this is really similar to `reconnectionDelay`.  Is there a name that better represents refreshing/reestablishing an existing connection that can be used here?

Perhaps 'reattemptPrimaryConnectionDelay'.  Something else?  I don't know, just brainstorming.  It's minor.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -486,6 +560,24 @@ public Socket createSocket() throws IOException {
         
     }
 
+    /**
+     * An immutable representation of a host and port
+     */
+    private static class HostInfo {
+    	private String host;
```
philsttr: Make `host` and `port` final to make this immutable.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -621,9 +769,17 @@ public void setPort(int port) {
     public int getPort() {
         return port;
     }
+    
+    public void setHostAddresses(String hostAddresses) {
```
philsttr: I'm debating whether I like a comma separated list of addresses, or a first-class list better.

For example, you could have:
```
/*
 * Add the given "hostname:port" to the list of possible destination hosts to which to send logs.
 * Hosts will be tried in the order in which they are added.
 * Logs will be sent to the first one which accepts the connection.
 */
void addHostAddress(String hostAddress)
```

And then have your logback config be this:
```
<appender....>
    <hostAddress>hostname1:port1</hostAddress>
    <hostAddress>hostname2:port2</hostAddress>
</appender>
```

Having separate `hostAddress` elements looks better to me, as long as you're not building up a list via variables in xml.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -375,6 +421,18 @@ private synchronized void openSocket() {
                     
                     this.socket = tempSocket;
                     this.outputStream = tempOutputStream;
+                    
+                    /*
+                     * If connected to a secondary, remember when the connection should be closed to
+                     * force attempt to reconnect to primary
+                     */
+                    if( hostIndex != 0 ) {
```
philsttr: _nitpick_: This would read a little better (and be more self-documenting) if you used a constant, or created a wrapper method.

`if (hostIndex != PRIMARY_HOST_INDEX) {`

or 

`if (isPrimaryHost(hostIndex)) {`

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -506,23 +598,80 @@ public synchronized void start() {
...
+	        			errorCount++;
+	        			addError("Invalid hostAddress '"+hostAndPort+"': unparseable value (expected format 'host:port')");
+	        			continue;
+	            	}
+	        		
+	        		String host = StringUtils.trim(parts[0]);
+	                try {
+	                    InetAddress.getByName(host);
+	                } catch (UnknownHostException ex) {
+	                    addError("Invalid hostAddress '"+hostAndPort+"': host unkown (was '"+host+"')");
```
philsttr: _type-o_: "unknown"

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -506,23 +598,80 @@ public synchronized void start() {
...
+	        if( hostAddresses != null ) {
+	        	String[] hostAndPorts = StringUtils.split(hostAddresses, ",");
+	        	
+	        	hostInfo = new HostInfo[hostAndPorts.length];
+	        	int hostIndex = 0;
+	        	
+	        	for(String hostAndPort: hostAndPorts) {
+	        		String[] parts = StringUtils.split(hostAndPort,':');
+	        		
+	        		if( parts.length != 2 ) {
```
philsttr: Maybe use the `port` field value if the port is not provided?

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -506,23 +598,80 @@ public synchronized void start() {
...
+	                     continue;
+	                }
+	        		if( port <= 0 ) {
+						addError("Invalid hostAddress '"+hostAndPort+"': port must be greater than 0 (was "+port+")");
+						errorCount++;
+						continue;
+	                }
+	                
+	                hostInfo[hostIndex] = new HostInfo(host, port);
+	                hostIndex++;
```
philsttr: _nitpick_: This is a personal preference thing, but I generally prefer using Lists for cases like this.  But I could go either way.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -506,23 +598,80 @@ public synchronized void start() {
...
+	        	}
+	        	
+	        	if( port > 0 || remoteHost != null ) {
```
philsttr: The default port is > 0, so this will always trigger a warning.

Just check if `remoteHost != null` here.

philsttr: The default port is > 0, so this will always print a warning.

Just check if `remoteHost != null` here

philsttr:
This looks really good!  Definitely worth integrating.

I made some (mostly minor) comments.

Will need unit tests and documentation updates to go along with it.

Also please remove tabs, and use spaces for indentation.

It will probably take me a while to get to this, so if you want to speed things up, feel free to update the PR.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
brenuart: Having one entry per host in the XML configuration looks for better - I agree. However, it makes things more difficult in some cases like Docker container environment.
It would be nice if we could set a single `LOG_ADDRESSES` environment variable with the addresses of log collector hosts in order and simply refer to it in the Logback configuration. Having separate entries per host would require the declare one environment variable for each as well. Do you see the point?

What do you think if we do support both schemes. User would shave three options:
1/ `<remoteHost>`, `<remotePort>`: legacy parameters
2/ collection of `<hostAddress>`: your proposed approach
3/ `<hostAddresses>`: comma separated list

These options would be exclusive.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
brenuart: Disabling this feature would mean the appender would stay on the connected host until a failure occurs. In other words, it wouldn't try to automatically reconnect on the primary. Some people may want that behaviour indeed...

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
brenuart: +1 for me as well.

brenuart:
Documentation is not yet updated - waiting for your feedback regarding the `<hostAddresses>` configuration parameter.
I'm also wondering if we shouldn't make it easier for people to extend the current implementation. We could make `TcpSendingEventHandler` protected instead of private for instance and make sure most interesting configuration parameters are available to child classes. If it had been the case, I could have written my own extension, in my own private java package with additional features you may not want to integrate in the code base.

I'll be off until 20th Augustus - hope you will have some time before then to review the changes and send feedback. Would be nice if the failover feature could be merged before 1st september and a new version released at that time.

philsttr:
Thanks so much!

Yes, I can take a look at making it more extensible as you mentioned.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
philsttr: Honestly, I've never really liked the name "host" / "port", because it doesn't indicate what is the "host" of  (source/destination?).  How about we do this:

```
/*
 * Add the given comma separated list of destination addresses ("hostname:port")
 * to the list of possible destinations to which to send logs.
 * Hosts will be tried in the order in which they are added.
 * Logs will be sent to the first one which accepts the connection.
 */
void addDestination(String destination)
```

And then the logback config be this:
```
<appender....>
    <destination>hostname1:port1, hostname2:port2</destination>
</appender>
```

OR this:
```
<appender....>
    <destination>hostname1:port1<destination>
    <destination>hostname2:port2</destination>
</appender>
```

And then we can deprecate the host/port fields.

brenuart: Great! I like it as well. 
I was a bit worried about deprecating the host/port fields... We could still keep them and make the two options exclusive. Maybe with a warning at runtime and in the doc saying the host/port will be removed in future releases.

philsttr: Yeah, that's what I meant.  Deprecated means the functionality keeps working, but usage is discouraged.

brenuart:
Work not yet finished. I made a quick commit with your last comments so you can have a look while I'm on holiday (until 21/8).
A few other comments:
- couldn't make it compile because of Java 1.7 construct (compiler is configured for source and target 1.6...) - sorry, no more time to fix it
- I would love to extract the code responsible to parse the host addresses in a separate utility class - would be easier to test
- wondering if wouldn't be a good idea to align the UDPAppender configuration so <destination> could also be used - would make both appender similar
- the reconnect logic is not very efficient in case of "short" network problem: the appender will immediately switch to a secondary if any, or pause for a relative long time (if only one destination) before attempting to reconnect. Maybe an exponential back off strategy would be better. Pluggable?

philsttr:
I've merged this PR into the `failover` branch.  As soon as I update the documentation and do some more testing, I'll merge it to master and release it.

I've changed the thread name logic when multiple destinations are used.  The thread names will now change based on the currently connected destination.

I've extracted the destination string parsing logic out into `DestinationParser`

I've changed the UnknownHostException to just be a warn at configuration/startup time.  The destination will still be added in case it is a transient problem.

Regarding UDPAppender and reconnection logic, I'd rather separate those out into separate issues/PRs so they can potentially be released in a future release.

Documentation has still not been updated.  I'll get to it soon.

philsttr:
Documentation updated on failover branch.

I've tested the branch locally, but would appreciate it if you give it a shot.

philsttr:
Hi @brenuart,

I will be unavailable 25/8-30/8, but I can potentially release this on 24th or 31st.

I'm happy with the changes, but I would ideally like confirmation from you that this works as you expect.

Let me know, and I'll get it released as soon as I can.

Phil

brenuart:
I'll have a look this morning. Expect a confirmation in the next few hours. Nice if we could release it today.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -506,25 +599,40 @@ public synchronized void start() {
             errorCount++;
             addError("No encoder was configured for appender " + name + ".");
```
brenuart: Some but not all error report the appender name in the message.
As far as I can tell, this is not needed as logback core already includes enough information about the failing appender when throwing its exception.
I propose to remove the appender name from every error messages generated by the appender.

brenuart:
Everything works as expected. Nothing against a release of the current codebase asis and delay fixes for the last few comments until next release.

philsttr:
Released in 4.5

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
philsttr: Can you open a new issue for this?  Or better yet, submit PR

Issue #117, opened 2015-07-31
Code requires Java 1.7 - pom configured to expect 1.6


brenuart:
The `pom.xml` configures the compiler to expect java source code in 1.6:

```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>2.3.2</version>
                <configuration>
                    <source>1.6</source>
                    <target>1.6</target>
                </configuration>
            </plugin>
```

However, the code base makes use of methods available starting from Java 1.7. Consequence is the project cannot be built with a JDK1.6 and requires at least 1.7 at runtime (even if the generated byte code is 1.6 compliant). One such method is:

```
AsyncDisruptorAppender.java:353: Undefined reference: void java.util.concurrent.ScheduledThreadPoolExecutor.setRemoveOnCancelPolicy(boolean)
```

Conclusions are:
- pom should require at least java 1.7
- compiler should be configured accordingly
- `animal-sniffer-maven-plugin` should be used to verify compliance at build time

(or the offending classes refactored to remove dependencies on java 1.7)

philsttr:
Yeah, I've tried to maintain compatibility with 1.6 as long as possible.  There were some complaints from users the last time a dependency on 1.7 was inadvertently introduced.

Currently, the code will run on a 1.6 JVM.  As you mentioned, the only 1.7+ api used (that I know of) is `ScheduledThreadPoolExecutor.setRemoveOnCancelPolicy`.  This method is guarded such that it will only be invoked at runtime if it exists.  See [here](https://github.com/logstash/logstash-logback-encoder/blob/287f1b77838d7fbf01080f691374ee68c7aed5e8/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java#L338-L347).

So, the code must be compiled with a 1.7+ jvm, but the bytecode produced in 1.6 format will run on a 1.6 jvm.  No 1.7+ apis are called when running on a jvm < 1.7.

The `animal-sniffer-maven-plugin` could be introduced if it has the ability to exclude certain occurrences that have been handled manually like above.

philsttr:
Configured animal-sniffer-plugin to check for apis incompatible with 1.6.

Excluded existing usage of `ScheduledThreadPoolExecutor.setRemoveOnCancelPolicy`, since that is handled properly at runtime.

philsttr:
Released in 4.5

PR #118, opened 2015-07-31
gh117 - Java version

brenuart:
Configure pom to require java 1.7, verify suitable JDK is used and validate code against 1.7


philsttr:
Not merging this PR.  See comments in  #117

Issue #121, opened 2015-07-31
Unable to set KeepAliveCharset via log back configuration


brenuart:
Unable to set `keepAliveCharset` via XML configuration. The following configuration fragment causes an error when log back configures itself:

```
<appender name="STASH" class="net.logstash.logback.appender.LogstashTcpSocketAppender">
    <keepAliveCharset>UTF-8</keepAliveCharset>
    ...
</appender>
```

Logback complains with the following error:
```
Exception in thread "main" java.lang.IllegalStateException: java.lang.IllegalStateException: Logback configuration error detected: 
ERROR in ch.qos.logback.core.joran.spi.Interpreter@38:24 - no applicable action for [keepAliveCharset], current ElementPath  is [[configuration][appender][keepAliveCharset]]
```

Apparently Logback doesn't recognise this parameter as a valid Java Bean property. This is probably caused by the fact that the corresponding getter and setter do not have the same type:

- `setKeepAliveCharset(String): void`: takes as String as input
- `getKeepAliveCharset(): Charset`: returns a `Charset`

philsttr:
PR #122 merged

PR #122, opened 2015-07-31
Fix KeepAliveCharset configuration property (gh121)

brenuart:
Let logback/joran handles the string/Charset conversion and make sure getter/setter are of the same type so the property is correctly recognised by Joran during configuration.
See https://github.com/logstash/logstash-logback-encoder/issues/121 for more information.


philsttr:
Thanks!  I'll test and get this merged as soon as I can.

PR #149, opened 2016-02-12
Don't log stacktrace on unknown host exception

stianl:
IMHO the stack trace is of limited value here. We're using destinations that will never resolve in DNS that's only used in local development (e.g. references to docker container names). 

Other warnings emitted on initializing logback typically don't include stack traces, e.g:

`13:16:49,461 |-WARN in ch.qos.logback.classic.LoggerContext[default] - Resource [logback.xml] occurs multiple times on the classpath.`


philsttr:
Thanks!

PR #155, opened 2016-03-02
Update logback

firone:
Update of logback version to 1.1.6 (to be compatible with last version of spring-boot-starter-logging)


philsttr:
Addressed in commit 37b157b195efb2a30487e322a4603fadd9cf3e04

Issue #139, opened 2015-11-17
LogstashAccessEncoder does not support global custom fields


avallen:
Would be good to point this out in the documentation which is not clear about this.

I wonder why this is and if the suggested alternative is to configure a `AccessEventCompositeJsonEncoder` together with a Pattern JSON provider.

philsttr:
Hi @avallen ,

It's just an oversight.

Yes, the current way to do this would be to use a pattern JSON provider.

You can add a pattern provider to a `LogstashAccessEncoder` like this:

```xml
<encoder class="net.logstash.logback.encoder.LogstashAccessEncoder">
    <provider class="net.logstash.logback.composite.accessevent.AccessEventPatternJsonProvider">
        <pattern>{"foo":"bar"}</pattern>
    </provider>
</encoder>
```

Or you can use `AccessEventCompositeJsonEncoder` instead of a `LogstashAccessEncoder` like you mentioned.

harbulot:
@philsttr I was unable to get the configuration you suggested in your previous comment to work. However, the PR (#157) I've just submitted adds support for these fields with `<customFields />`. It seems to work.

philsttr:
PR #157 merged to master.  Will be included in next release.

PR #157, opened 2016-03-08
Adding global custom fields to access encoder.

harbulot:
`LogstashAccessEncoder` did not support global custom fields (see issue #139). This patch copies the logic in `LogstashFormatter` into `LogstashAccessFormatter`, making use of a `GlobalCustomFieldsJsonProvider<IAccessEvent>`. The following configuration then works:

        <customFields>{"foo": "bar"}</customFields>
    </encoder>


philsttr:
Thanks for the contribution!

Issue #125, opened 2015-08-24
LogstachTcpSocketAppender: Roundrobin multiple IPs for single hostname


brenuart:
It would be nice if the appender had an option to consider every IP addresses that might be assigned to a single hostname and behave as if multiple destinations were manually configured.

Activation of the feature should be optional.
If disabled, the default resolution would apply, ie one of the IP is selected "at random".
If enabled, all IPs are added as destinations. Their ordering (primary vs. secondaries) is undetermined however...

philsttr:
Closing this because I'm not aware of an easy way to fetch all the IPs associated with a single hostname from java.  Feel free to reopen with a suggested way of doing so.  Or a PR.

PR #195, opened 2016-12-29
RandomDestination at first & round-robin

withccm:
Hello.
Logstash-logback-encoder is woking by primary and secondary.
I need a balanced request to each destination.
@brenuart registered the issue #125 

* The first set destination is random.
* If the size is exceeded, round-robin occurs. 

I hope to review the code.
Thanks.

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Formatter;
 import java.util.List;
+import java.util.Random;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
@@ -113,7 +114,7 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
      * Index into {@link #destinations} of the "primary" destination.
      */
     private static final int PRIMARY_DESTINATION_INDEX = 0;
-    
+
     /**
      * The host to which to connect and send events
      */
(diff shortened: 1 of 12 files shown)


src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -252,7 +253,32 @@
...
+     */
+    private Random firstRandomIndexGenerator = new Random();
+
+    /**
+     * If the size is exceeded, round-robin occurs.
+     * Destination cycles by round-robin.
+     *
+     * When null (the default), disabled round-robin.
+     */
+    private Integer roundRobinSize;
```
philsttr: I don't really understand why a size is needed.  It seems like you would want to either round robbin, or not.  Why would you want to _not_ round robin for a while, and then start round robin?

philsttr:
I think we need to introduce a `DestinationSelectionStrategy` concept.

I can see three strategies at the moment:
1. `PreferPrimary` - the current strategy of always preferring the primary destination if it is available  (`secondaryConnectionTTL` is only applicable for this strategy)
2. `RoundRobin` - a new strategy round-robin connects to a destination (starting at the first) every time a reconnect occurs
3. `Random` - randomly selects a destination every time a reconnect occurs

Would those satisfy your use cases?

From the changes in the current PR, it is not obvious (without studying the code pretty hard) which strategy is being used for any given set of configuration values.   For example, how does `secondaryConnectionTTL` come into play with your new strategies?  What happens when it is set?  So, I think a first-class `DestinationSelectionStrategy` concept is needed in order to make it very explicit what strategy is being used.

brenuart:
+1
That would be very nice indeed.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
withccm: As you know, if you do round robin, a reconnection occurs.
When reconnecting each time, a big cost was introduced, and the "roundRobinSize" was introduced as a concept of buffer.

withccm:
@philsttr 
I will modify it using the concept of the strategy you have told me.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
philsttr: Ohhhh,  I thought it would just round-robin the next time a connection was needed (i.e. if a connection was broken).  I didn't notice that your change forces a reconnect after every N log messages.  Makes sense now.  If this is the case, I think it would make more sense to use a time based round-robin, rather than message count based.  That way sudden increases of load don't trigger a reconnection storm.

This is kind of similar to the `secondaryConnectionTTL` already implemented.  However, since it applies to any connection.  So, maybe it would just be `connectionTTL`.

withccm:
@philsttr 
Hello. I have developed three strategies. Please check again.

philsttr:
Hey @withccm,

I've refactored your code a bit to extract a strategy interface, with three implementations.  This pulls the logic out the appender, and allows people to develop custom strategies.

I've committed it to the [destination_connection_strategy](https://github.com/logstash/logstash-logback-encoder/tree/destination_connection_strategy) branch.  I still have more testing and documentation to do, but this will provide a preview of how I'm thinking this should work.

withccm:
@philsttr, Thank you for your positive review.
I have reviewed the code you have been working on.
I would be willing to help if I needed anything.

philsttr:
4.10 released with this change

withccm:
Thanks @philsttr

Issue #208, opened 2017-02-28
#asJson


tetv:
I would like to see support for `#asJson{...}` to transform a string to json.
Same concept as the already existing `#asLong{...}` and `#asDouble{...}`

philsttr:
Merged #211

PR #211, opened 2017-03-01
Implementation of feature #asJson

tetv:
Hi, I have added support for using #asJson issue: #208 
`#asJson` also supports conversion out of the box to any primitive type like **int**, **double** or **boolean**.
It's useful specially when using with the **message** field:

```xml
<encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
    <providers>
        <pattern><pattern>{
            "bytes_sent_str": "%b",
            "bytes_sent_long": "#asJson{%b}",
            "has_message": "#asJson{%X{hasMessage}}",
            "raw_message": "#asJson{%m}"
        }</pattern></pattern>
    </providers>
</encoder>
```
...will produce something like:

```
{
  "bytes_sent_str": "1024",
  "bytes_sent_long": 1024,
  "has_message": true,
  "raw_message": { "type":"example", "msg":"example of json message with type" }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1605,7 +1605,9 @@ So this example...
 <pattern>
   {
     "bytes_sent_str": "%b",
-    "bytes_sent_long": "#asLong{%b}"
+    "bytes_sent_long": "#asLong{%b}",
+    "has_message": "#asJson(%X{hasMessage})",
+    "raw_message: "#asJson(%m)"
   }
 </pattern>
 ```
@@ -1614,7 +1616,9 @@ So this example...
 ```
 {
   "bytes_sent_str": "1024",
-  "bytes_sent_long": 1024
+  "bytes_sent_long": 1024,
+  "has_message": true,
+  "raw_message": { "type":"example", "msg":"example of json message with type" }
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -1597,6 +1597,7 @@ The operations are:
 
 * `#asLong{...}` - evaluates the pattern in curly braces and then converts resulting string to a long (or a null if conversion fails).
 * `#asDouble{...}` - evaluates the pattern in curly braces and then converts resulting string to a double (or a null if conversion fails).
+* `#asJson{...}` - evaluates the pattern in curly braces and then converts resulting string to json (or a null if conversion fails).
 
 So this example...
 
--- a/README.md
+++ b/README.md
@@ -1606,8 +1606,8 @@ So this example...
   {
     "bytes_sent_str": "%b",
     "bytes_sent_long": "#asLong{%b}",
-    "has_message": "#asJson(%X{hasMessage})",
-    "raw_message: "#asJson(%m)"
+    "has_message": "#asJson{%X{hasMessage}}",
+    "raw_message": "#asJson{%m}"
   }
 </pattern>
 ```
--- a/src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
+++ b/src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
@@ -19,6 +19,8 @@ import ch.qos.logback.core.spi.ContextAware;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1613,7 +1616,9 @@ So this example...
 ```
 {
   "bytes_sent_str": "1024",
-  "bytes_sent_long": 1024
+  "bytes_sent_long": 1024,
+  "has_message": true,
+  "raw_message": { "type":"example", "msg":"example of json message with type" }
```
philsttr: Where do `type` and `msg` come from in this example?    I don't understand how `#asJson{%m}` turns into `{ "type":"example", "msg":"example of json message with type" }`

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -160,6 +199,19 @@ protected Double transform(final String value)  throws NumberFormatException {
...
+        private static final ObjectMapper mapper = new ObjectMapper()
+                .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
+                .enable(JsonGenerator.Feature.ESCAPE_NON_ASCII);
+
+        public AsJsonValueTransformer(final ValueGetter<String, Event> generator) {
+            super(generator);
+        }
+
+        protected JsonNode transform(final String value) throws IOException {
+            return mapper.readTree(value);
```
philsttr: Hmm.  I'm wondering if this could be refactored slightly differently to avoid the json parsing.  I mean, as I understand it,  you ultimately just want to write the value to the generator as-is, right?.  e.g. `generator.writeRaw(String)`.  

The problem with the current design, is that `ComputableValueWriter` always calls `writeObject`.  So, if you actually wanted to call `writeRaw` instead of `writeObject`, you'd probably need to remove the `ValueGetter` concept, and just do the value retrieval, transformation, and writing in a `NodeWriter`.

In addition, I'm a little worried about the static configuration of the `ObjectMapper` used to parse the value.   i.e. what happens if this default configuration doesn't work for somebody's JSON data.  If parsing could be avoided, then there would be no need for the `ObjectMapper`.

Your thoughts?

philsttr:
As mentioned in the in-line comment, I think what is really needed is an `#asRaw` operation, since you are essentially pre-formatting the string as JSON.  

i.e. if the pattern evaluates to `{"foo":"bar"}`, you really just want to write `{"foo":"bar"}` to the json output.  There's no real need to parse it into a `JsonNode` and then serialize it back out as `{"foo":"bar"}`, because that's the same thing you started with.

Let me know if my understanding is incorrect.

tetv:
Hi @philsttr, thank you for your input.

The idea beyond `#asJson` is similar to `#asLong`, and `#asLong` validates if the `%message` is really a long, therefore in some how, there should be a way to validate if the `%message` that `#asJson` receives is a proper json message, else we can end up generating non valid json files.

`#asRaw` can be good to have as well, but assumes who is creating logs, creates always valid and wellformed json messages, which can be trick specially when you are using external libraries that generates important logs that you don't want to discard.

Happy, if there are better ways to validate if json is well formatted without using `ObjectMapper`.

Currently in production we are using a variation function `#tryJson` which try to convert to json (using `ObjectMapper`), and if not possible, then the `%message` is kept as is (json string). Therefore we never loose the `%message` and at same time the json logs are always valid.

Finally, is always good to have a human log message, but `slf4j` is a bit limited in terms of passing `StructuredArguments` therefore the way we found to pass them (`StructuredArguments`) is via a  json message. In the near future we would like to have something like `#asJson{%m, type}` in order to end up moving important fields to the root.

PS: I know that your library supports `StructuredArguments` however, in terms of architecture we don't want to have in our code any references to backlog or specific backlog implementation.

philsttr:
Hey @tetv,

This will be available in the next release.

I made one minor change... I made asJson reuse the jsonfactory (and its object mapper) that the encoder uses, rather than create a new object mapper.  This allows users to configure the jsonFactory if needed to customize the parsing.

mgosk:
@philsttr When do you prepare next release ? It will be convinient for me if You release version with this function soon.

philsttr:
I'll get it released this week

philsttr:
[4.10](https://github.com/logstash/logstash-logback-encoder/releases/tag/logstash-logback-encoder-4.10) released with this change

mgosk:
Thanks

03.06.2017 7:23 AM "Phil Clay" <notifications@github.com> napisał(a):

PR #228, opened 2017-05-12
Stack hash feature

### LINKED ISSUE #226: Compute hash value from exception stacks as a JSON field
Hi,

First of all, I'd like to thank you for this splendid project, I'm a big fan !

I'm using it a lot at Orange, and have developped some additional features that - I believe - are of pretty good interest.
See: https://github.com/Orange-OpenSource/orange-mathoms-logging

The most valuable one is probably the idea of generating a single hash from exception stacks.
Those hashes help matching several occurrences of the same error, as it is highly probable that two errors with the same hash will be - indeed - occurrences of the same error.

Once indexed by Elasticsearch, you can track your errors history, frequency, and build awesome dashboards...

I've just proposed contributing to Logback (see https://jira.qos.ch/browse/LOGBACK-1296) and don't know what will be the outcome, but it could also simply be a feature of logstash-logback-encoder as the full power of it is really related to Elasticsearch.

Let me know...

pismy:
closes #226 

- ShortenedThrowableConverter can now compute and inline hashes in the stack trace with the `inlineHash` param (`false` by default)
- StackHashJsonProvider makes available the topmost error hash as a separate JSON field (not a default field)

### DIFF
--- a/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
@@ -24,6 +24,7 @@ import net.logstash.logback.stacktrace.StackHasher;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
 
@@ -62,6 +63,18 @@ public class StackHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEve
 
     public void addExclude(String exclusionPattern) {
         excludes.add(Pattern.compile(exclusionPattern));
+    }
+
+    /**
+     * Set exclusion patterns as a list of coma separated patterns
+     * @param comaSeparatedPatterns list of coma separated patterns
+     */
+    public void setExclusions(String comaSeparatedPatterns) {
+        if (comaSeparatedPatterns == null || comaSeparatedPatterns.isEmpty()) {
+            this.excludes = new ArrayList<Pattern>(5);
+        } else {
+            setExcludes(Arrays.asList(comaSeparatedPatterns.split("\\s*\\,\\s*")));
+        }
     }
 
     public void setExcludes(List<String> exclusionPatterns) {
--- a/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
@@ -1,0 +1,97 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.logstash.logback.composite.loggingevent;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.IThrowableProxy;
+import ch.qos.logback.classic.spi.ThrowableProxy;
+import com.fasterxml.jackson.core.JsonGenerator;
+import net.logstash.logback.composite.AbstractFieldJsonProvider;
+import net.logstash.logback.composite.JsonWritingUtils;
+import net.logstash.logback.stacktrace.StackElementFilter;
+import net.logstash.logback.stacktrace.StackHasher;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * A JSON provider that adds a {@code stack_hash} Json field on a log with a stack trace
+ * <p>
+ * This hash is computed using {@link StackHasher}
+ * 
+ * @author Pierre Smeyers
+ */
+public class StackHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
+
+    public static final String FIELD_NAME = "stack_hash";
+
+    /**
+     * Patterns used to determine which stacktrace elements to exclude.
+     *
+     * The strings being matched against are in the form "fullyQualifiedClassName.methodName"
+     * (e.g. "java.lang.Object.toString").
+     *
+     * Note that these elements will only be excluded if and only if
+     * more than one consecutive line matches an exclusion pattern.
+     */
+    private List<Pattern> excludes = new ArrayList<Pattern>(5);
+
+    private StackHasher hasher = new StackHasher();
+
+    public StackHashJsonProvider() {
+        setFieldName(FIELD_NAME);
+    }
+
+    @Override
+    public void start() {
+        if(!excludes.isEmpty()) {
+            hasher = new StackHasher(StackElementFilter.byPattern(excludes));
+        }
+        super.start();
+    }
+
+    public void addExclude(String exclusionPattern) {
+        excludes.add(Pattern.compile(exclusionPattern));
+    }
+
+    public void setExcludes(List<String> exclusionPatterns) {
+        if (exclusionPatterns == null || exclusionPatterns.isEmpty()) {
+            this.excludes = new ArrayList<Pattern>(5);
+        } else {
+            this.excludes = new ArrayList<Pattern>(exclusionPatterns.size());
+            for (String pattern : exclusionPatterns) {
+                addExclude(pattern);
+            }
+        }
+    }
+
+    public List<String> getExcludes() {
+        List<String> exclusionPatterns = new ArrayList<String>(excludes.size());
+        for (Pattern pattern : excludes) {
+            exclusionPatterns.add(pattern.pattern());
+        }
+        return exclusionPatterns;
+    }
+
+    @Override
+    public void writeTo(JsonGenerator generator, ILoggingEvent event) throws IOException {
+        IThrowableProxy throwableProxy = event.getThrowableProxy();
+        if (throwableProxy != null && throwableProxy instanceof  ThrowableProxy) {
+        	String hash = hasher.hexHash(((ThrowableProxy)event.getThrowableProxy()).getThrowable());
+            JsonWritingUtils.writeStringField(generator, getFieldName(), hash);
+        }
+    }
+}
--- a/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import net.logstash.logback.composite.AbstractFieldJsonProvider;
 import net.logstash.logback.composite.JsonWritingUtils;
 import net.logstash.logback.stacktrace.StackElementFilter;
-import net.logstash.logback.stacktrace.StackHasher;
+import net.logstash.logback.stacktrace.ThrowableHasher;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -30,7 +30,7 @@ import java.util.regex.Pattern;
 /**
  * A JSON provider that adds a {@code stack_hash} Json field on a log with a stack trace
  * <p>
- * This hash is computed using {@link StackHasher}
+ * This hash is computed using {@link ThrowableHasher}
  * 
  * @author Pierre Smeyers
  */
@@ -49,7 +49,7 @@ public class StackHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEve
      */
     private List<Pattern> excludes = new ArrayList<Pattern>(5);
 
-    private StackHasher hasher = new StackHasher();
+    private ThrowableHasher hasher = new ThrowableHasher();
 
     public StackHashJsonProvider() {
         setFieldName(FIELD_NAME);
@@ -58,7 +58,7 @@ public class StackHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEve
     @Override
     public void start() {
         if(!excludes.isEmpty()) {
-            hasher = new StackHasher(StackElementFilter.byPattern(excludes));
+            hasher = new ThrowableHasher(StackElementFilter.byPattern(excludes));
         }
         super.start();
     }
--- a/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import net.logstash.logback.composite.AbstractFieldJsonProvider;
 import net.logstash.logback.composite.JsonWritingUtils;
 import net.logstash.logback.stacktrace.StackElementFilter;
-import net.logstash.logback.stacktrace.ThrowableHasher;
+import net.logstash.logback.stacktrace.StackHasher;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -28,15 +28,15 @@ import java.util.List;
 import java.util.regex.Pattern;
 
 /**
- * A JSON provider that adds a {@code error_hash} Json field on a log with a stack trace
+ * A JSON provider that adds a {@code stack_hash} Json field on a log with a stack trace
  * <p>
- * This hash is computed using {@link ThrowableHasher}
+ * This hash is computed using {@link StackHasher}
  * 
  * @author Pierre Smeyers
  */
-public class ErrorHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
+public class StackHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
 
-    public static final String FIELD_NAME = "error_hash";
+    public static final String FIELD_NAME = "stack_hash";
 
     /**
      * Patterns used to determine which stacktrace elements to exclude from hash computation.
@@ -46,16 +46,16 @@ public class ErrorHashJsonProvider extends AbstractFieldJsonProvider<ILoggingEve
      */
     private List<Pattern> excludes = new ArrayList<Pattern>(5);
 
-    private ThrowableHasher hasher = new ThrowableHasher();
+    private StackHasher hasher = new StackHasher();
 
-    public ErrorHashJsonProvider() {
+    public StackHashJsonProvider() {
         setFieldName(FIELD_NAME);
     }
 
     @Override
     public void start() {
         if(!excludes.isEmpty()) {
-            hasher = new ThrowableHasher(StackElementFilter.byPattern(excludes));
+            hasher = new StackHasher(StackElementFilter.byPattern(excludes));
         }
         super.start();
     }
--- a/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
+++ b/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
@@ -13,10 +13,7 @@
  */
 package net.logstash.logback.stacktrace;
 
-import java.util.ArrayList;
-import java.util.Deque;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
 
@@ -612,7 +609,19 @@ public class ShortenedThrowableConverter extends ThrowableHandlingConverter {
     public void addExclude(String exclusionPattern) {
         excludes.add(Pattern.compile(exclusionPattern));
     }
-    
+
+    /**
+     * Set exclusion patterns as a list of coma separated patterns
+     * @param comaSeparatedPatterns list of coma separated patterns
+     */
+    public void setExclusions(String comaSeparatedPatterns) {
+        if (comaSeparatedPatterns == null || comaSeparatedPatterns.isEmpty()) {
+            this.excludes = new ArrayList<Pattern>(5);
+        } else {
+            setExcludes(Arrays.asList(comaSeparatedPatterns.split("\\s*\\,\\s*")));
+        }
+    }
+
     public void setExcludes(List<String> exclusionPatterns) {
         if (exclusionPatterns == null || exclusionPatterns.isEmpty()) {
             this.excludes = new ArrayList<Pattern>(5);
--- a/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
+++ b/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
@@ -161,7 +161,7 @@ public class ShortenedThrowableConverter extends ThrowableHandlingConverter {
 
     private StackElementFilter stackElementFilter;
 
-    private ThrowableHasher throwableHasher;
+    private StackHasher stackHasher;
 
     /**
      * Evaluators that determine if the stacktrace should be logged.
@@ -173,13 +173,21 @@ public class ShortenedThrowableConverter extends ThrowableHandlingConverter {
         parseOptions();
         // instantiate stack element filter
         if(excludes == null || excludes.isEmpty()) {
-            stackElementFilter = StackElementFilter.any();
+            if(inlineHash) {
+                // filter out elements with no source info
+                addInfo("[inlineHash] is active with no exclusion pattern: use non null source info filter to exclude generated classnames (see doc)");
+                stackElementFilter = StackElementFilter.withSourceInfo();
+            } else {
+                // use any filter
+                stackElementFilter = StackElementFilter.any();
+            }
         } else {
+            // use patterns filter
             stackElementFilter = StackElementFilter.byPattern(excludes);
         }
         // instantiate stack hasher if "inline hash" is active
         if(inlineHash) {
-            throwableHasher = new ThrowableHasher(stackElementFilter);
+            stackHasher = new StackHasher(stackElementFilter);
         }
         super.start();
     }
@@ -259,7 +267,7 @@ public class ShortenedThrowableConverter extends ThrowableHandlingConverter {
         // compute stack trace hashes
         Deque<String> stackHashes = null;
         if(inlineHash && (throwableProxy instanceof ThrowableProxy)) {
-            stackHashes = throwableHasher.hexHashes(((ThrowableProxy) throwableProxy).getThrowable());
+            stackHashes = stackHasher.hexHashes(((ThrowableProxy) throwableProxy).getThrowable());
         }
 
         /*
@@ -597,8 +605,8 @@ public class ShortenedThrowableConverter extends ThrowableHandlingConverter {
         this.inlineHash = inlineHash;
     }
 
-    protected void setThrowableHasher(ThrowableHasher throwableHasher) {
-        this.throwableHasher = throwableHasher;
+    protected void setStackHasher(StackHasher stackHasher) {
+        this.stackHasher = stackHasher;
     }
 
     public void addExclude(String exclusionPattern) {
--- a/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
+++ b/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
@@ -14,6 +14,7 @@
 package net.logstash.logback.stacktrace;
 
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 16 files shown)


pismy:
For now, I didn't dare to activate stack hashes by default in ShortenedThrowableConverter, or to make StackHashJsonProvider a standard field

Maybe we can *incubate* it before doing so.

pismy:
Well, I may have a bug on root cause first and suppressed errors.

I will Check and fix it next week.

pismy:
Indeed, I had an issue with suppressed errors.
Fixed and squashed

I chose not to compute stack hash for suppressed errors (doesn't seem very useful), but my choice is challengeable ;-)

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
```diff
@@ -0,0 +1,120 @@
...
+        // skip null element, generated class or filter element
+        return element != null && element.getFileName() != null && element.getLineNumber() >= 0 && filter.accept(element);
+    }
+
+    int hash(StackTraceElement element) {
+        int result = element.getClassName().hashCode();
+        result = 31 * result + element.getMethodName().hashCode();
+        // let's assume filename is not necessary
+        result = 31 * result + element.getLineNumber();
+        return result;
```
philsttr: Why not just use element.hashCode()?

src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
```diff
@@ -0,0 +1,97 @@
...
+            exclusionPatterns.add(pattern.pattern());
+        }
+        return exclusionPatterns;
+    }
+
+    @Override
+    public void writeTo(JsonGenerator generator, ILoggingEvent event) throws IOException {
+        IThrowableProxy throwableProxy = event.getThrowableProxy();
+        if (throwableProxy != null && throwableProxy instanceof  ThrowableProxy) {
+        	String hash = hasher.hexHash(((ThrowableProxy)event.getThrowableProxy()).getThrowable());
```
philsttr: This hash value only takes the top-most throwable into consideration, correct?

It seems like it would be more beneficial to take the complete chain of throwables into account when computing the hash here.  That would increase the chance that a unique hash value represents a unique error condition.

I'm thinking of the case where there is some  code that catches downstream exceptions, and wraps them in a new exception with a generic message.  Since this hash value is only taking the top-most throwable into consideration, there would be a wide range of root causes that would generate the same hash value.

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
```diff
@@ -230,16 +248,22 @@ public String convert(ILoggingEvent event) {
         if (throwableProxy == null || isExcludedByEvaluator(event)) {
             return CoreConstants.EMPTY_STRING;
         }
-        
+
+        // compute stack trace hashes
+        Deque<String> stackHashes = null;
```
philsttr: Rather than calculating all the hashes up front, and creating an intermediary `Deque` object, I think it would be simpler (and more efficient) to calculate hashes of individual throwables as the converter is walking through the cause chain.

i.e Don't calculate the hashes here.  Don't create a Deque object.  Just call the StashHasher in `appendFirstLine`.  In this case, StashHasher would have a method that calculates the hash of the throwable passed in as an argument without taking its causes into consideration.

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
```diff
@@ -150,7 +152,14 @@
      * True to print the root cause first.  False to print exceptions normally (root cause last).
      */
     private boolean rootCauseFirst;
-    
+
+    /**
+     * True to compute and inline stack hashes.
+     */
+    private boolean inlineHash;
```
philsttr: Should provide a way to set this to true via parsed options.  See the `parseOptions` method.

src/main/java/net/logstash/logback/stacktrace/StackElementFilter.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package net.logstash.logback.stacktrace;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * Component in charge of accepting or rejecting {@link StackTraceElement elements} when computing a stack trace hash
+ */
+public abstract class StackElementFilter {
```
philsttr: Would be nice to make `ShortenedThrowableConverter.isIncluded` use this as well.   Right now the logic is duplicated.

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
```diff
@@ -0,0 +1,120 @@
...
+        hash(error, hexHashes);
+        return hexHashes;
+    }
+
+    /**
+     * Generates a hash (int) of the given error stack
+     * <p>
+     * Two errors with the same stack hash are most probably same errors
+     */
+    int hash(Throwable error, Deque<String> hexHashes) {
```
philsttr: Do you want to consider the throwable's message in the hash?

Seems like throwables with different messages should generate different hashes, no?

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+/**
+ * Utility class that generates a hash for any Java {@link Throwable error} stack
+ *
+ * @author Pierre Smeyers
+ */
+public class StackHasher {
```
philsttr: Think I would name this `ThrowableHasher` since the input methods take a Throwable as an argument (and not a list of StackElements)

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
```diff
@@ -0,0 +1,120 @@
...
+        return hexHashes(error).peek();
+    }
+
+    /**
+     * Generates and returns Hexadecimal hashes for the error stack and each ancestor {@link Throwable#getCause() cause}
+     * <p>
+     * The first queue element is the stack hash for the topmost error, the next one (if any) is it's direct
+     * {@link Throwable#getCause() cause} hash, and so on...
+     */
+    public Deque<String> hexHashes(Throwable error) {
```
philsttr: As I alluded to in other comments, I don't think StackHasher should have a method that returns a `Deque`.  The Deque complicates the code a bit (since callers need to know how to correlate entries with causes), and will result in extra garbage in a code "hot path".

Instead, I would make it provide a method like this:

```java
/**
 * Generates and returns Hexadecimal hash for the given Throwable.
 *
 * @param throwable The throwable for which to calculate the hash
 * @param includeCauses when true, each cause will be considered when calculating the hash.
 *                      when false, only the top level throwable will be considered in the hash.
 */
public String computeHash(Throwable throwable, boolean includeCauses)
```


StackHashJsonProvider would call this method once, using includeCauses=true, since it only needs one value for the complete throwable chain.

ShortenedThrowableConverter would call this method using includeCauses=false for each throwable in the cause chain as it recursively traverses each throwable.

src/test/java/net/logstash/logback/composite/loggingevent/StackHashJsonProviderTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testDefaultName() throws IOException {
+        // GIVEN
+        when(event.getThrowableProxy()).thenReturn(throwableProxy);
+        when(throwableProxy.getThrowable()).thenReturn(new Exception("test error"));
+        // WHEN
+        provider.writeTo(generator, event);
+        // THEN
+        verify(generator).writeStringField(eq(StackHashJsonProvider.FIELD_NAME), anyString());
```
philsttr: Maybe verify that the arg can be parsed as a hex integer?

src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
philsttr: Nevermind, I didn't that StackHasher recurses through each cause when calculating the hash for a throwable.

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
```diff
@@ -159,6 +168,15 @@
     @Override
     public void start() {
         parseOptions();
+        // instantiate stack hasher if need be
+        if(inlineHash) {
+            // use same exclusion patterns to compute hash
+            if(excludes == null || excludes.isEmpty()) {
+                stackHasher = new StackHasher();
+            } else {
+                stackHasher = new StackHasher(StackElementFilter.byPattern(excludes));
```
philsttr: Should the StackHasher take the excludes into consideration?  The excludes are mainly there to shorten log output.  It seems like for error correlation you would always want to include all stack elements.  Your thoughts?

pismy:
Hi @philsttr 

Thanks for your comments, very constructive I appreciate.

For better understanding, I probably have to explain better what I'm trying to do with this feature.

Actually the `stack_hash` is meant to identify an error (throwable) with a **short** and **stable** *identifier* (or *signature*, I don't really know how to call it):

* **short** for easying Elasticsearch indexing, and take advantage of it: that's why I chose a hash (hex encoded)
* **stable** is the tricky part, as the same type of error occurring twice may not generate exactly the same stack trace

Let's consider error 1:
<pre>
com.xyz.MyApp$MyClient$MyClientException: <b>An error occurred while getting Alice's things</b>
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:26)
  at com.xyz.MyApp$MyService.displayThings(MyApp.java:16)
  <b>at com.xyz.MyApp$MyService$$FastClassByCGLIB$$e7645040.invoke()</b>
  <i>at net.sf.cglib.proxy.MethodProxy.invoke()
  at org.springframework.aop.framework.Cglib2AopProxy$CglibMethodInvocation.invokeJoinpoint()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed()
  at sun.reflect.NativeMethodAccessorImpl.invoke0()
  at sun.reflect.NativeMethodAccessorImpl.invoke()
  at sun.reflect.DelegatingMethodAccessorImpl.invoke()
  at java.lang.reflect.Method.invoke()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod()
  at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.AbstractTraceInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.transaction.interceptor.TransactionInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.framework.Cglib2AopProxy$DynamicAdvisedInterceptor.intercept()</i>
  ...
Caused by: com.xyz.MyApp$HttpStack$HttpError: <b>I/O error on GET http://dummy/user/alice/things</b>
  at com.xyz.MyApp$HttpStack.get(MyApp.java:40)
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:24)
  ... 23 common frames omitted
Caused by: java.net.SocketTimeoutException: Read timed out
  at com.xyz.MyApp$HttpStack.get(MyApp.java:38)
  ... 24 common frames omitted
</pre>

Elements in **bold** my vary from one occurrence to the other:

*  error message often contains stuff related to the very error occurrence context,
* AOP and dynamic invocation / generated classes may vary from one execution to another

Elements in *italic* are somewhat not stable. Ex:

* JEE container stuff: may change when you upgrade your JEE container version or add/remove/reorganize your servlet filters chain for instance
* Spring Framework underlying stacks (MVC, security) for pretty much the same reason

Now let's consider error 2:
<pre>
com.xyz.MyApp$MyClient$MyClientException: <b>An error occurred while getting Bob's things</b>
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:26)
  at com.xyz.MyApp$MyService.displayThings(MyApp.java:16)
  <b>at com.xyz.MyApp$MyService$$FastClassByCGLIB$$07e70d1e.invoke()</b>
  <i>at net.sf.cglib.proxy.MethodProxy.invoke()
  at org.springframework.aop.framework.Cglib2AopProxy$CglibMethodInvocation.invokeJoinpoint()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed()
  at sun.reflect.NativeMethodAccessorImpl.invoke0()
  at sun.reflect.NativeMethodAccessorImpl.invoke()
  at sun.reflect.DelegatingMethodAccessorImpl.invoke()
  at java.lang.reflect.Method.invoke()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod()
  at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.AbstractTraceInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.transaction.interceptor.TransactionInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.framework.Cglib2AopProxy$DynamicAdvisedInterceptor.intercept()</i>
  ...
Caused by: com.xyz.MyApp$HttpStack$HttpError: <b>I/O error on GET http://dummy/user/bob/things</b>
  at com.xyz.MyApp$HttpStack.get(MyApp.java:40)
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:24)
  ... 23 common frames omitted
Caused by: java.net.SocketTimeoutException: Read timed out
  at com.xyz.MyApp$HttpStack.get(MyApp.java:38)
  ... 24 common frames omitted
</pre>

You may see in this example that the **bold** elements are all different from error 1.

Still it is the same exact error (despite the context is different as it applies to another user), and I want to be able to count them as two occurrences of the same error.

Now let's consider error 3:
<pre>
com.xyz.MyApp$MyClient$MyClientException: <b>An error occurred while getting Alice's things</b>
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:26)
  at com.xyz.MyApp$MyService.displayThings(MyApp.java:16)
  <b>at com.xyz.MyApp$MyService$$FastClassByCGLIB$$e7645040.invoke()</b>
  <i>at net.sf.cglib.proxy.MethodProxy.invoke()
  at org.springframework.aop.framework.Cglib2AopProxy$CglibMethodInvocation.invokeJoinpoint()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed()
  at sun.reflect.NativeMethodAccessorImpl.invoke0()
  at sun.reflect.NativeMethodAccessorImpl.invoke()
  at sun.reflect.DelegatingMethodAccessorImpl.invoke()
  at java.lang.reflect.Method.invoke()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs()
  at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod()
  at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.AbstractTraceInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.transaction.interceptor.TransactionInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke()
  at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed()
  at org.springframework.aop.framework.Cglib2AopProxy$DynamicAdvisedInterceptor.intercept()</i>
  ...
Caused by: com.xyz.MyApp$HttpStack$HttpError: <b>I/O error on GET http://dummy/user/alice/things</b>
  at com.xyz.MyApp$HttpStack.get(MyApp.java:40)
  at com.xyz.MyApp$MyClient.getTheThings(MyApp.java:24)
  ... 23 common frames omitted
Caused by: <b>javax.net.ssl.SSLException: Connection has been shutdown: javax.net.ssl.SSLHandshakeException: Received fatal alert: certificate_unknown</b>
        at com.sun.net.ssl.internal.ssl.SSLSocketImpl.checkEOF(SSLSocketImpl.java:1172)
  ... 24 common frames omitted
</pre>

Here, you can see that the first and second errors are the same as in error 1, but the root cause is different (`SSLException` instead of `SocketTimeoutException`).

So in that case I don't want the top error hash to be the same.

As a conclusion, I apply the following rules:

1. an error hash shall not compute against the error message
2. an error hash shall compute against it's parent cause (recurses)
3. in order to stablize the error hash (over time and space), it's nice to be able to strip the stack trace from non-stable elements

This said, I will now answer your comments, refering to those rules.

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
pismy: Yes I could do that... I chose to ignore the filename part

No real reason for that. What's your opinion ?

src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java
pismy: Yes. That's rule 2 in my comment:

2. an error hash shall compute against it's parent cause (recurses)

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
pismy: Considering rule 2, we cannot compute error hashes individually.

We could change the `Deque` design, but that would compell to (re)compute several times the root cause errors hash...

Let's discuss about the most appropriate design

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
pismy: You're right.

I saw the `parseOptions` thing, but don't really understand how it works (didn't find any doc about it)

But I'll give it a look

src/main/java/net/logstash/logback/stacktrace/StackElementFilter.java
pismy: You're right

I didn't dare to refactor too much of your code ;-)

But I will

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
pismy: That's rule 1 (see below)

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
pismy: Very good idea. I'll do it

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
pismy: Based on my comments, let's find out the best design

src/test/java/net/logstash/logback/composite/loggingevent/StackHashJsonProviderTest.java
pismy: right

src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java
pismy: Well, that's a very good question related to rule 3 (see below) that we could rephrase as:

"Does the exclusion patterns for outputing the stack trace match the exclusion patterns for computing the error hashes ?"

Not 100% sure about the answer, but in general case I would say **yes**.

What do you think ?

pismy:
BTW I'm a bit of a newbie in GitHub, and I don't know how you prefer working with PR

Do you want me to keep squashing my commits, or do you want to keep the history as several commits ?

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
pismy: What do you think, shall I rename `StackHashJsonProvider` as `ThrowableHashJsonProvider` ?

(with `stack_hash` default property moved to `error_hash` ?)

src/main/java/net/logstash/logback/stacktrace/StackElementFilter.java
pismy: One question:

I see in your impl that you use [Pattern.find()](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#find())

But in your README examples of patterns, it looks like you could use [Pattern.match()](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#match()) instead (probably better in terms of performances):

```xml
  <throwableConverter class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
    <maxDepthPerThrowable>30</maxDepthPerThrowable>
    <maxLength>2048</maxLength>
    <shortenedClassNameLength>20</shortenedClassNameLength>
    <exclude>sun\.reflect\..*\.invoke.*</exclude>
    <exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
    <evaluator class="myorg.MyCustomEvaluator"/>
    <rootCauseFirst>true</rootCauseFirst>
  </throwableConverter>
</encoder>
```

I guess changing the impl from `find()` to `match()` may screw up some existing projects using your library, but maybe that could be wise to mention it in the doc and present some optimized patterns (using `^` and `$`):

```xml
  <throwableConverter class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
    <maxDepthPerThrowable>30</maxDepthPerThrowable>
    <maxLength>2048</maxLength>
    <shortenedClassNameLength>20</shortenedClassNameLength>
    <exclude>^sun\.reflect\..*\.invoke.*</exclude>
    <exclude>^net\.sf\.cglib\.proxy\.MethodProxy\.invoke$</exclude>
    <evaluator class="myorg.MyCustomEvaluator"/>
    <rootCauseFirst>true</rootCauseFirst>
  </throwableConverter>
</encoder>
```

src/main/java/net/logstash/logback/stacktrace/ThrowableHasher.java
```diff
@@ -0,0 +1,120 @@
...
+        this(StackElementFilter.any());
+    }
+
+    /**
+     * Generates a Hexadecimal hash for the given error stack
+     * <p>
+     * Two errors with the same stack hash are most probably same errors
+     */
+    public String hexHash(Throwable error) {
+        return hexHashes(error).peek();
```
philsttr: Maybe optimize this usage to not require the Deque

src/main/java/net/logstash/logback/stacktrace/ThrowableHasher.java
```diff
@@ -0,0 +1,120 @@
...
+
+        // push hexadecimal representation of hash
+        hexHashes.push(String.format("%08x", hash));
+
+        return hash;
+    }
+
+    boolean accept(StackTraceElement element) {
+        // skip null element, generated class or filter element
+        return element != null && element.getFileName() != null && element.getLineNumber() >= 0 && filter.accept(element);
```
philsttr: I think the `element.getFileName() != null && element.getLineNumber() >= 0` check will cause classes that weren't generated with debug information to not be evaluated. i.e. only classes compiled with debug info have the file/line number info.

If you're ok with that, then that's fine.  But probably worth documenting.

If you're not ok with that, then you'll need to remove the check, and just rely on the user to configure the filter to remove dynamically generated stack frames.

src/main/java/net/logstash/logback/stacktrace/StackElementFilter.java
philsttr: Good suggestion.  Yeah, probably too late to change `match()` to `find()` now.  Will update docs

src/main/java/net/logstash/logback/stacktrace/StackHasher.java
philsttr: Hmm.  I'm a little conflicted on this myself.

On one hand, the Hasher takes a `Throwable`, and uses the Throwable's class and stack to compute the hash.  So `ThrowableHasher` makes sense.

On the other hand, the field that contains the throwable is named `stack_trace`.

So, I could go either way.  Naming is hard.  ;)

I think I like `stack_hash` and `StackHashJsonProvider` to be parallel with `stack_trace` and `StackTraceJsonProvider`.  Which makes me want to go back to `StackHasher` as well.   ¯\\_(ツ)_/¯

philsttr:
Thanks for the detailed explanation.  It definitely cleared up my understanding of what you're trying to do.  The design decisions you made, and the resulting code, make perfect sense now.

And thank you for addressing my other comments.

I'm satisfied with your current design, including the Deque.  (with one comment to optimize the `hexHash` method to not require the Deque)

As for several commits vs squashing.  I don't usually squash unless a crazy refactoring was done during the lifetime of the branch.  I don't think the changes you've made qualify for that.  So just keep layering on commits.

pismy:
I'm glad my explainations made things clearer.
But that makes me think this feature probably deserves to be well documented, and I'm not really sure actual documentation is enough.
Should I write a dedicated chapter in the README ? Or at least enhance the doc ?

You have to know that I'm currently developping a complete playground demo application (microservices, Spring Boot, Spring Cloud, ELK, and obviously `logstash-logback-encoder`).
This project should be available on GitHub in the coming weeks, and will have extensive documentation, tutorials, ...
Maybe we could just make a link between the 2 projects to *get more info* ?

I'll work on your 2 comments today.

Thanks again.

src/main/java/net/logstash/logback/stacktrace/ThrowableHasher.java
pismy: You're right, I would simply have missed this case.

The beauty of peer review ;-)

pismy:
Ooh sorry I didn't see your comment about renaming `stack_hash` to `error_hash`
(actually I liked too `stack_hash` being related to `stack_trace`, that's why I called it this way in the first place)
I just commited the change, but I may revert it if you prefer
Just tell me

pismy:
Hey @philsttr 
Not directly related to this PR, but I've just released the demo app I was talking you about.
Check-out this: https://github.com/Orange-OpenSource/woofer

I'll migrate my private "stack hash" impl to logstash-logback-encoder as soon as available.

pismy:
Hi @philsttr 
I just published a detailed page about error hash computation in my former library project (based on our conversation)
https://github.com/Orange-OpenSource/orange-mathoms-logging/blob/master/stack-hash.md
Let me know if it might be useful to add something similar to logstash-logback-encoder along with this PR...

philsttr:
Sorry for the delay.

Love the sample app.  Nice job.

I think we can finalize this PR if the following are addressed:

1. Let's rename everything back to Stack*   (`StashHasher`, `stack_hash`, etc), so its parallel with stacktrace.  Sorry for the flip-flop.  It's hard to get naming right, but I think this is best now.
2. I like your detailed explanation on the hash computation.  Can you add that file to this PR, and link to it from the readme?

pismy:
No prob' about flip-flop on naming: I agree with you and I like taking some time to think. Our discussion was very fruitful.

I'll work on this next week. Can't wait to have it released !

BTW I changed my mind about default ThrowableHasher configuration if no exclusion pattern is specified.
Using a filter that accepts all StackTraceElements will not filter out generated classes, and the computed hash will never be stable.
So in that case I prefer using a default filter that filters out StackTraceElements with no source information (null filename or line numer < 0)
Too bad for classes not compiled in debug mode, but I will write a clear explanation in the doc about this.
Okay for you ?

philsttr:
Yeah. That's fine. If it becomes a problem, we can make it configurable in the future. Keep it simple for now.

pismy:
Hey, 
My latest commit takes into account all your comments and suggestions. 
I'm happy with it now.

philsttr:
Will take a look and get it merged soon.  Thanks!

pismy:
Hey,
Sorry, one more change that seemed quite important.
`StackHashJsonProvider` and `ShortenedThrowableConverter` now support a new `<exclusions>` element to set all exclusion patterns at once (as a coma separated list)
This was necessary to share same exclusion patterns and even make it configurable from outside Logback config with a variable (details and examples in stack-hash.md)

philsttr:
Merged.  Will be available in the next release.

I released 4.10 last week, and I try to have a few features / bug fixes in each release.  So, will probably wait a few more weeks until the next release.

Thanks again for this work!

PR #244, opened 2017-08-11
Fix typos - ssl tags

vgalda:


philsttr:
Not quite sure I understand this change.

`<ssl/>` is the same as `<ssl></ssl>` from an [xml standpoint](https://www.w3schools.com/xml/xml_elements.asp).

vgalda:
Oh, my bad. I didnť know it. Close it

PR #256, opened 2017-11-27
TurboFilter which puts Exception class names into fields

jackhammer2k:
add TurboFilter which puts the Throwable class name (actual and root cause) into fields so they can be listed/searched easily in ELK

### DIFF
--- a/src/main/java/net/logstash/logback/turbo/ThrowableAsField.java
+++ b/src/main/java/net/logstash/logback/turbo/ThrowableAsField.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


philsttr:
Thanks for the PR @jackhammer2k!

Since your goal is to provide additional fields in the JSON output, and since filters are supposed to be for deciding if a logging event should be logged, I think this would be better implemented as a [`JsonProvider`](https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/composite/JsonProvider.java) rather than a filter.  All of the other fields in the JSON output are provided via various implementations of `JsonProvider`, so you can use any of them as a reference for your implementation.

Also, you might be interested in the [`stackHash`](https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/composite/loggingevent/StackHashJsonProvider.java) provider for correlating errors.

PR #257, opened 2017-11-28
add Throwable class name as field

jackhammer2k:
add Provider which puts the Throwable class name (actual and root cause) into fields so they can be listed/searched easily in ELK


jackhammer2k:
I would like to activate it as default provider in LogstashFormatter, because its very helpful to get an overview of occurred error types. Do you agree?

philsttr:
Thanks @jackhammer2k . This will be available in the next release.

I'm hesitant to add new fields to the default LogstashFormatter, as that could affect a lot of people.  Existing users can  enable it if they need it.

PR #268, opened 2018-02-19
master: Added post socket connect hook for post authorization

hannes-angst:
I implemented a logmet (lumberjack protocol) encoder wrapper and appender for IBM Bluemix that relies heavily on logstash's marvelous work. 
There is however one obstacle I want to remove with this humple pull request. IBM/softlayer's security does not rely on certificates and trust, but on a API key which must be past after the socket connection was established. The very good implementation of `AbstractLogstashTcpSocketAppender` does not allow extension of a after connection callback which I so dearly need.
I hope you will accept this pull request, since it does not change the logic in any way and would allow me use logstash out of the box instead of shadowing the `AbstractLogstashTcpSocketAppender` class.

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
@@ -562,6 +562,8 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                         updateCurrentThreadName();
                     }
                     
+                    afterSocketOpen(socket);
+
                     this.readerFuture = scheduleReaderRunnable(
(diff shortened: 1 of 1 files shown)


philsttr:
Hi @hannes-angst 

I'm going to close this in favor of the new listener support being added in 5.0.

In 5.0, you will be able to register a listener that will be notified when a new connection is established.  The listener has access to the underlying `Socket`, so it should be able to perform the custom steps you need.

https://github.com/logstash/logstash-logback-encoder/blob/376d4b75d9028bc40a24b1586fa9a1138e5a1b78/src/main/java/net/logstash/logback/appender/listener/TcpAppenderListener.java#L55

hannes-angst:
Works like a charm :)

PR #283, opened 2018-04-11
Added more packages to OSGI's exports

redvasily:
This is necessary for structured arguments or creating custom field types.
Just a simple change to OSGi Manifest file


philsttr:
Thanks!

Issue #279, opened 2018-03-29
Configurable newline symbol in ShortenedThrowableConverter


fred84:
Log consumer in our system use `\n` as message separator.
Stack trace elements are concatenated by [CoreConstants.LINE_SEPARATOR](https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java#L505), so traces are split across several lines.
Currently I found solution proposed in https://github.com/logstash/logstash-logback-encoder/issues/83 , but I think It is worth adding an option to configure newline symbol inside json. If you don't see and obstacles, I can create PR.

philsttr:
Seems reasonable to me.  A PR would be great!

philsttr:
Hi @fred84 

Thank you so much for your contribution!

I have merged it, made a few naming changes, and documented it in the README.md.  It will be available in 5.1 when it is released.

I made the naming changes because I wanted to use consistent terminology with Jackson.  For example, Jackson uses `CharacterEscapes` and `escapeSequence`, so I wanted logstash-logback-encoder to use the same.

The one functional change I made was to start with the standard ASCII escaping behavior by default, rather than starting with ESCAPE_NONE for every ASCII character.  If your use case requires that you start with ESCAPE_NONE for every character, then you can set `<includeStandardAsciiEscapesForJSON>false</includeStandardAsciiEscapesForJSON>`

PR #285, opened 2018-04-17
Configurable char replacement JsonFactoryDecorator 

fred84:
Solves #279 . Configuration example:
         
```
                <jsonFactoryDecorator class="net.logstash.logback.decorate.ReplaceCharsJsonFactoryDecorator">
                    <replace>
                        <targetNumber>10</targetNumber> <!-- \n symbol -->
                        <replacement>===</replacement>
                    </replace>
                </jsonFactoryDecorator>
            </encoder>
```

### DIFF
--- a/src/main/java/net/logstash/logback/decorate/ReplaceCharsJsonFactoryDecorator.java
+++ b/src/main/java/net/logstash/logback/decorate/ReplaceCharsJsonFactoryDecorator.java
@@ -5,43 +5,48 @@ import com.fasterxml.jackson.core.io.CharacterEscapes;
 import com.fasterxml.jackson.core.io.SerializedString;
 import com.fasterxml.jackson.databind.MappingJsonFactory;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


fred84:
@philsttr Should I also add configuration example to README?

src/main/java/net/logstash/logback/decorate/ReplaceCharsJsonFactoryDecorator.java
```diff
@@ -0,0 +1,131 @@
...
+public class ReplaceCharsJsonFactoryDecorator implements JsonFactoryDecorator {
+
+    private static class CustomizableCharacterReplacer extends CharacterEscapes {
+
+        private final int[] escapeCodesForAscii;
+        private final List<Replace> replacements = new ArrayList<>();
+
+        private CustomizableCharacterReplacer() {
+            escapeCodesForAscii = new int[128];
+            Arrays.fill(escapeCodesForAscii, ESCAPE_NONE);
```
philsttr: I think this should start with `CharacterEscapes.standardAsciiEscapesForJSON()` by default rather than full of `ESCAPE_NONE`, mainly because the `standardAsciiEscapesForJSON()` are used by default if this `JsonFactoryDecorator` is not used.

i.e. If you happen to configure the `JsonFactoryDecorator` with no escapes, the behavior should be the same as if you didn't configure the `JsonFactoryDecorator` at all.

PR #286, opened 2018-04-24
Unix timestamp support for LogstashEncoder and LogstashAccessEncoder

peterjurkovic:
In our projects, we need to generate `@timestamp` as a unix timestamp but is not quite easy to do that with current implementation.  I suggest adding a new configuration property for `LogstashEncoder` and `LogstashAccessEncoder`

Example

```xml
  <unixTimestamp>true</unixTimestamp>
</encoder>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1156,8 +1156,14 @@ You can change the timezone like this:
 ```
 
 The value of the `timeZone` element can be any string accepted by java's  `TimeZone.getTimeZone(String id)` method.
+It is also possible generate `@timestamp` as a unix timestamp using following configuration:
 
-
+```
+<encoder class="net.logstash.logback.encoder.LogstashEncoder">
+  <unixTimestamp>true</unixTimestamp>
+</encoder>
+```
+An output will looks like `"@timestamp":"1524597357413"`.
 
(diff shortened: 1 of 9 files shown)


philsttr:
Should the unix timestamp be a numerical value instead of a string?

peterjurkovic:
It should be a numeric value but there are some cases where it is needed as a string (I would need it to be a string).

I was also thinking ... Instead of using `<unixTimestamp>` element we could introduce special pattern for `FormattedTimestampJsonProvider` where a user would pass e.g. `[UNIX_TIMESTAMP]` and would be generated the timestamp. 

e.g.

```
<pattern>[UNIX_TIMESTAMP]</pattern>
```
output 
```
"@timestamp":1524597357413
```

```
<pattern>[STRING_UNIX_TIMESTAMP]</pattern>
```
output 
```
"@timestamp":"1524597357413"
```

witch would cover both usecases. What do you think @philsttr ?

philsttr:
Hey @peterjurkovic .  I liked the "special pattern" approach.  I went ahead and implemented it.  It'll be in 5.1 when it is released.

Thank you for the suggestion.

PR #293, opened 2018-06-08
Add sequence number provider

kampka:
When configured, this provider outputs an incrementing sequence number for every log event.
This can be useful to detect gaps in your log stream and help identify problems with your log transport.
It also makes reading logs easier as it helps the reader to distinguish between missing log lines and flaky system behavior.

### DIFF
--- a/src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
@@ -1,0 +1,44 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.logstash.logback.composite.loggingevent;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import com.fasterxml.jackson.core.JsonGenerator;
+import net.logstash.logback.composite.AbstractFieldJsonProvider;
+import net.logstash.logback.composite.JsonWritingUtils;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Outputs an incrementing sequence number.
+ * Useful for determining if log events get lost along the transport chain.
+ */
+public class SequenceProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
+
+    public static final String FIELD_SEQUENCE = "sequence";
+
+    private static AtomicLong sequenceNumber = new AtomicLong(0L);
+
+    public SequenceProvider() { setFieldName(FIELD_SEQUENCE);}
+
+    @Override
+    public void writeTo(JsonGenerator generator, ILoggingEvent iLoggingEvent) throws IOException {
+        JsonWritingUtils.writeStringField(generator, getFieldName(), String.valueOf(sequenceNumber.incrementAndGet()));
+    }
+
+    void resetSequenceNumber() {
+        sequenceNumber.set(0L);
+    }
+}
--- a/README.md
+++ b/README.md
@@ -1183,7 +1183,7 @@ For example:
 
 ## Customizing JSON Factory and Generator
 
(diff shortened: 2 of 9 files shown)


src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final String FIELD_SEQUENCE = "sequence";
+
+    private static AtomicLong sequenceNumber = new AtomicLong(0L);
+
+    public SequenceProvider() { setFieldName(FIELD_SEQUENCE);}
+
+    @Override
+    public void writeTo(JsonGenerator generator, ILoggingEvent iLoggingEvent) throws IOException {
+        JsonWritingUtils.writeStringField(generator, getFieldName(), String.valueOf(sequenceNumber.incrementAndGet()));
```
philsttr: Would recommend writing this as a number field, instead of a string

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Outputs an incrementing sequence number.
+ * Useful for determining if log events get lost along the transport chain.
+ */
+public class SequenceProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
+
+    public static final String FIELD_SEQUENCE = "sequence";
+
+    private static AtomicLong sequenceNumber = new AtomicLong(0L);
```
philsttr: This should be `final`.

Also, did you intend to make this static, so that the sequence is global to the entire application?  It seems like this should be an instance field, so the `sequenceNumber` is per provider instance.  This would allow you to configure two different encoders (possibly to two different destinations), each with their own sequences, to keep track of them separately.

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public SequenceProvider() { setFieldName(FIELD_SEQUENCE);}
+
+    @Override
+    public void writeTo(JsonGenerator generator, ILoggingEvent iLoggingEvent) throws IOException {
+        JsonWritingUtils.writeStringField(generator, getFieldName(), String.valueOf(sequenceNumber.incrementAndGet()));
+    }
+
+    void resetSequenceNumber() {
```
philsttr: if `sequenceNumber` is static, this method should probably also be static to indicate that it is resetting the global static sequence.  However, if you change `sequenceNumber` to be an instance field, then leave this as an instance method.

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+import net.logstash.logback.composite.JsonWritingUtils;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Outputs an incrementing sequence number.
+ * Useful for determining if log events get lost along the transport chain.
+ */
+public class SequenceProvider extends AbstractFieldJsonProvider<ILoggingEvent> {
```
philsttr: I'd recommend naming this `SequenceJsonProvider` to follow the naming convention of all of the other `JsonProvider`s in this library.

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import com.fasterxml.jackson.core.JsonGenerator;
+import net.logstash.logback.composite.AbstractFieldJsonProvider;
+import net.logstash.logback.composite.JsonWritingUtils;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Outputs an incrementing sequence number.
```
philsttr: If you keep `sequenceNumber` as a static field (per my other comment), you should document that it is a global sequence, as opposed to an instance-level sequence

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
kampka: Was only needed for convenience in testing, hence removed.

src/main/java/net/logstash/logback/composite/loggingevent/SequenceProvider.java
kampka: I believe this no longer applies.

kampka:
Thank you for the review, I rebase the commit with your remarks applied.

philsttr:
Merged.  Thanks!

PR #294, opened 2018-06-08
Don't fill in stack trace in test exception

kampka:
When configured, this provider outputs an incrementing sequence number for every log event.
This can be useful to detect gaps in your log stream and help identify problems with your log transport.
It also makes reading logs easier as it helps the reader to distinguish between missing log lines and flaky system behavior.


philsttr:
Thanks!

PR #307, opened 2018-10-05
Add support for excluding markers via <includeTags>false</includeTags>

ramsperger:
This behavior mimics the includeMdc behavior and causes
the tags serialization to be skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,7 +762,8 @@ The field names can be customized (see [Customizing Standard Field Names](#custo
 | `level`       | String name of the level of the event
 | `level_value` | Integer value of the level of the event
 | `stack_trace` | (Only if a throwable was logged) The stacktrace of the throwable.  Stackframes are separated by line endings.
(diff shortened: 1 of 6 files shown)


philsttr:
Awesome!  Thanks.

BoinaRamesh:
@ramsperger/ @philsttr  I'm using LogstashEncoder for logging, recently i found there is tag which is populating in my logs. 

"marker": {
    "name": "LS_APPEND_OBJECT"
  }

using net.logstash.logback:logstash-logback-encoder:4.6

How can i suppress above properties(marker) from logging? does it work if i use <includeTags>?

ramsperger:
@BoinaRamesh - You need to set up a config for the encoder that includes `<includeTags>false</includeTags>`. 

Much abbreviated example:
```xml
<configuration>
    <appender name="stash" class="ch.qos.logback.core.rolling.RollingFileAppender">
            <includeTags>false</includeTags>
        </encoder>
    </appender>
</configuration>
```

PR #365, opened 2019-10-09
remove FastDateFormat

J-Jimmy:
Replace FastDateFormat with DateTimeFormatter and Instant from java.time package

### DIFF
--- a/src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
@@ -69,72 +69,7 @@ public abstract class FormattedTimestampJsonProvider<Event extends DeferredProce
     /**
      * Writes the timestamp to the JsonGenerator.
      */
-    private DateTimeFormatter timestampWriter = DateTimeFormatter.ofPattern(pattern, Locale.getDefault());
-
-    /**
-     * Writes the timestamp to the JsonGenerator
-     */
-    private static interface TimestampWriter {
-        void writeTo(JsonGenerator generator, String fieldName, long timestampInMillis) throws IOException;
-        
-        String getTimestampAsString(long timestampInMillis);
-    }
-    
-    /**
-     * Writes the timestamp to the JsonGenerator as a string formatted by the pattern.
-     */
-    private static class PatternTimestampWriter implements TimestampWriter {
-
-        private final DateTimeFormatter formatter;
-
-        public PatternTimestampWriter(DateTimeFormatter formatter) {
-            this.formatter = formatter;
-        }
-
-        
-        @Override
-        public void writeTo(JsonGenerator generator, String fieldName, long timestampInMillis) throws IOException {
-            JsonWritingUtils.writeStringField(generator, fieldName, getTimestampAsString(timestampInMillis));
-        }
-
-        @Override
-        public String getTimestampAsString(long timestampInMillis) {
-            return formatter.format(Instant.ofEpochMilli(timestampInMillis));
-        }
-    }
-    
-    /**
-     * Writes the timestamp to the JsonGenerator as a number of milliseconds since unix epoch.
-     */
-    private static class NumberTimestampWriter implements TimestampWriter {
-
-        @Override
-        public void writeTo(JsonGenerator generator, String fieldName, long timestampInMillis) throws IOException {
-            JsonWritingUtils.writeNumberField(generator, fieldName, timestampInMillis);
-        }
-        
-        @Override
-        public String getTimestampAsString(long timestampInMillis) {
-            return Long.toString(timestampInMillis);
-        }
-    }
-    
-    /**
-     * Writes the timestamp to the JsonGenerator as a string representation of the of milliseconds since unix epoch.
-     */
-    private static class StringTimestampWriter implements TimestampWriter {
-
-        @Override
-        public void writeTo(JsonGenerator generator, String fieldName, long timestampInMillis) throws IOException {
-            JsonWritingUtils.writeStringField(generator, fieldName, getTimestampAsString(timestampInMillis));
-        }
-        
-        @Override
-        public String getTimestampAsString(long timestampInMillis) {
-            return Long.toString(timestampInMillis);
-        }
-        
-    }
+    private DateTimeFormatter timestampWriter = DateTimeFormatter.ofPattern(pattern, Locale.getDefault()).withZone(timeZone != null ? timeZone.toZoneId() : TimeZone.getDefault().toZoneId());
     
     public FormattedTimestampJsonProvider() {
         setFieldName(FIELD_TIMESTAMP);
@@ -147,8 +82,7 @@ public abstract class FormattedTimestampJsonProvider<Event extends DeferredProce
     
     @Override
     public void writeTo(JsonGenerator generator, Event event) throws IOException {
-        String formattedTime = timestampWriter.format(Instant.ofEpochMilli(getTimestampAsMillis(event)));
-        generator.writeStringField(getFieldName(), formattedTime);
+        generator.writeStringField(getFieldName(), String.valueOf(getTimestampAsMillis(event)));
     }
 
     protected String getFormattedTimestamp(Event event) {
@@ -162,11 +96,11 @@ public abstract class FormattedTimestampJsonProvider<Event extends DeferredProce
      */
     private void updateTimestampWriter() {
         if (UNIX_TIMESTAMP_AS_NUMBER.equals(pattern)) {
-            timestampWriter = DateTimeFormatter.ofPattern("n");
+            timestampWriter = null;
         } else if (UNIX_TIMESTAMP_AS_STRING.equals(pattern)) {
-            timestampWriter = DateTimeFormatter.ofPattern(DEFAULT_PATTERN);
+            timestampWriter = DateTimeFormatter.ofPattern(DEFAULT_PATTERN).withZone(TimeZone.getDefault().toZoneId());
         } else {
-            timestampWriter = DateTimeFormatter.ofPattern(pattern, Locale.getDefault());
+            timestampWriter = DateTimeFormatter.ofPattern(pattern, Locale.getDefault()).withZone(TimeZone.getDefault().toZoneId());
         }
     }
     
--- a/src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
@@ -14,11 +14,12 @@
 package net.logstash.logback.composite;
 
 import java.io.IOException;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
 import java.util.TimeZone;
 
 import net.logstash.logback.fieldnames.LogstashCommonFieldNames;
-
-import org.apache.commons.lang3.time.FastDateFormat;
 
 import ch.qos.logback.core.spi.DeferredProcessingAware;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
```diff
@@ -160,11 +96,11 @@ protected String getFormattedTimestamp(Event event) {
      */
     private void updateTimestampWriter() {
         if (UNIX_TIMESTAMP_AS_NUMBER.equals(pattern)) {
-            timestampWriter = new NumberTimestampWriter();
+            timestampWriter = null;
```
philsttr: This is a backwards incompatible change that I won't be able to accept.

logstash-logback-encoder still needs to optionally support writing timestamps as numbers indicating the number of milliseconds since unix epoch

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
```diff
@@ -160,11 +96,11 @@ protected String getFormattedTimestamp(Event event) {
...
         } else if (UNIX_TIMESTAMP_AS_STRING.equals(pattern)) {
-            timestampWriter = new StringTimestampWriter();
+            timestampWriter = DateTimeFormatter.ofPattern(DEFAULT_PATTERN).withZone(TimeZone.getDefault().toZoneId());
```
philsttr: This is a backwards incompatible change that I won't be able to accept.

logstash-logback-encoder still needs to optionally support writing timestamps as strings indicating the number of milliseconds since unix epoch

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
```diff
@@ -160,11 +96,11 @@ protected String getFormattedTimestamp(Event event) {
...
         } else {
-            timestampWriter = new PatternTimestampWriter(FastDateFormat.getInstance(pattern, timeZone));
+            timestampWriter = DateTimeFormatter.ofPattern(pattern, Locale.getDefault()).withZone(TimeZone.getDefault().toZoneId());
```
philsttr: Since there are subtle differences in the pattern strings used by `DateTimeFormatter` and `FastDateFormat`, this is a backwards incompatible change.  I'm willing to accept this change, but this will need a major version bump of logstash-logback-encoder.

I'll probably hold off on this until logback supports timestamps with higher precision.

src/test/java/net/logstash/logback/encoder/LogstashAccessEncoderTest.java
```diff
@@ -200,7 +200,7 @@ public void unixTimestampAsNumber() throws Exception {
         
         JsonNode node = MAPPER.readTree(encoded);
         
-        assertThat(node.get("@timestamp").numberValue()).isEqualTo(timestamp);
+        assertThat(node.get("@timestamp").asLong()).isEqualTo(timestamp);
```
philsttr: Per my previous comment, this change will need to be reverted to ensure backwards compatibility.  The json node should be a number value, not a string value that needs to be converted to a Long

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -70,12 +66,12 @@
 public class LogstashEncoderTest {
     
     private static Logger LOG = LoggerFactory.getLogger(LogstashEncoderTest.class);
-    
+
```
philsttr: <nitpick> Would you be willing to revert all these whitespace changes?

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -107,43 +103,42 @@ public void basicsAreIncluded_logback12() throws Exception {
...
         JsonNode node = MAPPER.readTree(encoded);
-        
+
         verifyBasics(timestamp, node);
     }
 
     protected void verifyBasics(final long timestamp, JsonNode node) {
-        assertThat(node.get("@timestamp").textValue()).isEqualTo(FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSZZ").format(timestamp));
+        assertThat(node.get("@timestamp").textValue()).isEqualTo(String.valueOf(timestamp));
```
philsttr: I don't understand this change.  Can you clarify why the expected value is now a long value converted to a string, instead of a formatted timestamp string?

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -107,43 +103,42 @@ public void basicsAreIncluded_logback12() throws Exception {
...
         encoder.start();
         byte[] encoded = encoder.encode(event);
-        
+
         JsonNode node = MAPPER.readTree(encoded);
-        
-        assertThat(node.get("@timestamp").textValue()).isEqualTo(FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSZZ").format
-                (timestamp));
+
+        assertThat(node.get("@timestamp").textValue()).isEqualTo(String.valueOf(timestamp));
```
philsttr: Same comment as above.  Why is the expected value here a long value converted to a string?

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -156,34 +151,34 @@ public void basicsAreIncludedWithShortenedNames() throws Exception {
...
         byte[] encoded = encoder.encode(event);
-        
+
         String output = new String(encoded, "UTF-8");
-        
+
         assertThat(output).isEqualTo(String.format(
                 "{%n"
-                + "  @timestamp : \"" + FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSZZ").format(timestamp) + "\",%n"
+                + "  @timestamp : \"" + String.valueOf(timestamp) + "\",%n"
```
philsttr: Same comment as above. Why is the expected value here a long value converted to a string?

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -192,213 +187,212 @@ public JsonGenerator decorate(JsonGenerator generator) {
...
         encoder.setShortenedLoggerNameLength(length);
         encoder.start();
         byte[] encoded = encoder.encode(event);
-        
+
         JsonNode node = MAPPER.readTree(encoded);
 
-        assertThat(node.get("@timestamp").textValue()).isEqualTo(FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSZZ").format
-                (timestamp));
+        assertThat(node.get("@timestamp").textValue()).isEqualTo(String.valueOf(timestamp));
```
philsttr: Same comment as above. Why is the expected value here a long value converted to a string?

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
```diff
@@ -557,124 +551,123 @@ public void markerIsJSON() throws Exception {
...
         encoder.start();
         byte[] encoded = encoder.encode(event);
-        
+
         JsonNode node = MAPPER.readTree(encoded);
-        
-        assertThat(node.get("@timestamp").textValue()).isEqualTo(FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSZZ", TimeZone.getTimeZone("UTC")).format
-                (timestamp));
+
+        assertThat(node.get("@timestamp").textValue()).isEqualTo(String.valueOf(timestamp));
```
philsttr: Same comment as above. Why is the expected value here a long value converted to a string?

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
philsttr: This new code is always using the default timezone.   It needs to take the configured `timeZone` into consideration like the old code.

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
J-Jimmy: For this one, you have to take a close look at the overall change, or maybe I should remove UNIX_TIMESTAMP_AS_NUMBER completely because currently its only used in 2 test cases for it....
However, what was done previously, you would take the timestamp as milliseconds, convert them to a Date, and then take this NumberTimestampWriter to write it back as timestamp in milliseconds.

So actually im not using it anymore, I will remove it to clear confusion.
(Nothing lost here)

src/test/java/net/logstash/logback/encoder/LogstashAccessEncoderTest.java
J-Jimmy: Sorry, Long is a type of Number? Its just the correct one already.

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: I missunderstood the StringTimestampWriter, as a result I've done some wrong changes here

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: I missunderstood the StringTimestampWriter, as a result I've done some wrong changes here

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: I missunderstood the StringTimestampWriter, as a result I've done some wrong changes here

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: I missunderstood the StringTimestampWriter, as a result I've done some wrong changes here

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: I missunderstood the StringTimestampWriter, as a result I've done some wrong changes here

src/test/java/net/logstash/logback/encoder/LogstashEncoderTest.java
J-Jimmy: This one is annoying me aswell, I didnt had this change until I commited, something wrong in my Settings I guess, I will look into it.

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
J-Jimmy: I missunderstood the StringTimestampWriter, I will take a second look

src/test/java/net/logstash/logback/encoder/LogstashAccessEncoderTest.java
philsttr: They are not quite the same.

`numberValue()` will return the number if and only if the JsonNode represents a JSON number.  This is the expected condition of this test case.

`asLong()` will convert String JsonNode values to a number.  In this test case, the JsonNode is not expected contain a String value.

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
philsttr: We cannot remove the UNIX_TIMESTAMP_AS_NUMBER, or the `NumberTimestampWriter`.

The `FormattedTimestampJsonProvider` supports three scenarios:

UNIX_TIMESTAMP_AS_NUMBER will write the timestamp value as a json _number_ indicating the number of milliseconds since unix epoch, like this:

```
{
    "@timestamp" : 1570901563
}
```

UNIX_TIMESTAMP_AS_STRING will write the timestamp value as a json _string_ indicating the number of milliseconds since unix epoch, like this:

```
{
    "@timestamp" : "1570901563"
}
```

And any other pattern will be used as a format string.  e.g. 

```
{
    "@timestamp" : "2011-12-03T10:15:30+01:00"
}
```

From what I can tell, the new code always writes the timestamp value as a json _string_, which has effectively removed support for UNIX_TIMESTAMP_AS_NUMBER.

philsttr:
Closing this in favor of #377 

Thanks @J-Jimmy for filing the issue, and contributing the PR.  Even though I'm not accepting this PR directly, I used your contribution as a guide.  I'll definitely call out your effort in the release notes when this is released.

PR #377, opened 2019-11-23
Switch FastDateFormat to DateTimeFormatter

philsttr:
Switch usage of `FastDateFormat` to `DateTimeFormatter`

Remove dependency on apache commons-lang3 (since `FastDateFormat` was the only thing being used).

Switch default timestamp format to `DateTimeFormatter.ISO_OFFSET_DATE_TIME`

Added the ability to specify `DateTimeFormatter` constant names as patterns.

This change is NOT fully backwards compatible, since:
* the default timestamp pattern changed to `DateTimeFormatter.ISO_OFFSET_DATE_TIME`
* the pattern strings supported by `DateTimeFormatter` are not exactly the same as those supported previously by `FastDateFormat`

When this PR is merged, the major version will need to be bumped due to the backwards incompatibilities.
Therefore, holding off on merging this PR until ready for a new major release.


mfriess2:
We have the problem with the random sorted log messages, too.
Are there any plans on releasing this path soon?

ELK released the fix with 7.0? https://github.com/elastic/elasticsearch/issues/10005
However, Logback still seems to be unfinished, too? https://jira.qos.ch/browse/LOGBACK-1374

philsttr:
Hi @mfriess2 , this change to logstash-logback-encoder to switch to `DateTimeFormatter` will not help your situation.  Higher precision timestamps will need to be supported in logback (via the logback issue you mentioned) before logstash-logback-encoder can take advantage of them.  Since the pace of development of logback is extremely slow, and this would likely be a major change to logback, I am not expecting it to be implemented in logback any time soon.

PR #378, opened 2019-11-30
Generic masking support

philsttr:


smuryginim:
Hello! These gone be really useful features.
Can I ask if there is some feasible date for the improvement? I try to understand if it is ready in the nearest time, or I need to implement the same on my own?

philsttr:
@smuryginim I'll try to get it released this week.

smuryginim:
Thanks for the response! I cloned the repo and played a bit with the improvement. 
Really great that you provided custom extension points!

I had a case like: 
`{
    "filed1": "blabla",
    "message": "Log whole message body: {\"personalInfo\": \"Please mask me!\"}"
}`

Would it be a nice solution to implement ValueMasker and hide the required fields?

Another one feature, which can be useful (at least for my case, possibly it also can be useful for another case) is to add some kind of field to mask Provider, which can provide the list of required fields. The fields can be detected for some internal logic which is defined by a user (for my case the fields can be collected from the model classes, which are marked by some annotation).
What do you thing about it?

philsttr:
For your first question (masking fields within an embedded json string ), I can see two solutions:

1. Mask sensitive fields prior to logging it.  In other words, use the `MaskingJsonGenerator` when adding the embedded json to your log message.  e.g. `logger.log("Log whole message body {}", writeObjectWithMasking(body))` where `writeObjectWithMasking(body)` is a method you provide that writes using a`MaskingJsonGenerator`
2. A regex masker that looks for embedded json fields.  This would be a fairly expensive regex.

Do you have any other ideas?  I'd personally recommend 1 for the best performance.  2 will be pretty slow to perform for all logged messages.

For your second question (a fields to mask provider), that seems reasonable.  I've created #382

smuryginim:
I have the same doubts for a performance about the approach with ValueMasker.
For performance reasons, it will be better to go with approach 1.

bastey:
I'm faced the same situation as @smuryginim.
So, I'm trying to implement the 1st approch that @philsttr mentions by using the `MaskingJsonGenerator`.
> where `writeObjectWithMasking(body)` is a method you provide that writes using a`MaskingJsonGenerator`

Do you have somewhere an example of how using `MaskingJsonGenerator` ?
Message I want to mask some parts is like :
```
----------------------------------------------------------------------------------------
URI          : /TOTO/TITI
Method       : POST
Headers      : {content-type=[application/json]}
Request body : {"field1":"value1","field2":"value_i_want_to_mask","field3":value3}
----------------------------------------------------------------------------------------
```

smuryginim:
> I'm faced the same situation as @smuryginim.
> So, I'm trying to implement the 1st approch that @philsttr mentions by using the `MaskingJsonGenerator`.
> 
> > where `writeObjectWithMasking(body)` is a method you provide that writes using a`MaskingJsonGenerator`
> 
> Do you have somewhere an example of how using `MaskingJsonGenerator` ?
> Message I want to mask some parts is like :
> 
> ```
> ----------------------------------------------------------------------------------------
> URI          : /TOTO/TITI
> Method       : POST
> Headers      : {content-type=[application/json]}
> Request body : {"field1":"value1","field2":"value_i_want_to_mask","field3":value3}
> ----------------------------------------------------------------------------------------
> ```

I uploaded the whole source code, and there were a lot of examples in unit tests.
I hope that will help!

bastey:
Sorry @smuryginim but where ? :confused:

smuryginim:
You can download full git repository of the project, and then it ispossible to check class with simple find ussages.

philsttr:
Also check the [unit tests](https://github.com/logstash/logstash-logback-encoder/blob/master/src/test/java/net/logstash/logback/mask/MaskingJsonGeneratorDecoratorTest.java) for many usages of `MaskingJsonGeneratorDecorator`

PR #386, opened 2020-01-01
Avoid potential race condition in DeferredLogstashMarker#getSuppliedV…

PascalSchumacher:
…alue by not publishing a null LogstashMarker.

By the way: Thank you very much for developing `logstash-logback-encoder` and a happy new year!


philsttr:
I don't think this actually fixes any race conditions, since the changed code is within a synchronized block, but I'll merge it anyway, since it removes a volatile write.

PascalSchumacher:
On further thinking about it I belive you are correct.

It is not a race condition, because only `null` can be visible to other threads before the end of the synchronized block. And if the value is `null` they will block waiting for the lock.

Sorry for the missleading comment and pull request title. :-(

PR #391, opened 2020-02-26
Check that event has caller data

ArthurGazizov:
From 'getCallerData()' documentation "Note that calling this event may trigger the computation of caller data." To avoid computation can verify that event has CallerData

### DIFF
--- a/src/main/java/net/logstash/logback/composite/loggingevent/CallerDataJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/loggingevent/CallerDataJsonProvider.java
@@ -59,6 +59,9 @@ public class CallerDataJsonProvider extends AbstractFieldJsonProvider<ILoggingEv
         event.getCallerData();
     }
     private StackTraceElement extractCallerData(final ILoggingEvent event) {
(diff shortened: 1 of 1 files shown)


philsttr:
This pull request has failing unit tests.  Please fix the code, or the tests.

However, from looking at the [implementation of LoggingEvent](https://github.com/qos-ch/logback/blob/7ee74ad132bf74ead68c5bb553f5353aa2588bb0/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L277-L287)...

```java
    public StackTraceElement[] getCallerData() {
        if (callerDataArray == null) {
            callerDataArray = CallerData
                            .extract(new Throwable(), fqnOfLoggerClass, loggerContext.getMaxCallerDataDepth(), loggerContext.getFrameworkPackages());
        }
        return callerDataArray;
    }

    public boolean hasCallerData() {
        return (callerDataArray != null);
    }
```

... it looks as though calling getCallerData() is required in order for hasCallerData() to ever return true.

Therefore, I'm not sure if the call to hasCallerData() should be performed prior to calling getCallerData().

PR #393, opened 2020-03-03
Added isIncludeContext check to access logger encoder.

kersten:
This PR adds the same functionality to the access logger that is already implemented in the standard JSON logger.

If one set the element `includeContext` to `false` the context of the logger is not added to the JSON output.

```xml
<configuration>
    <!-- always a good activate OnConsoleStatusListener -->
    <statusListener class="ch.qos.logback.core.status.OnConsoleStatusListener"/>
    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
            <timeZone>UTC</timeZone>
            <includeContext>false</includeContext>
            <fieldNames>
                <requestHeaders>request_headers</requestHeaders>
                <responseHeaders>response_headers</responseHeaders>
            </fieldNames>
        </encoder>
    </appender>
    <appender-ref ref="STDOUT"/>
</configuration>
```

### DIFF
--- a/src/main/java/net/logstash/logback/LogstashAccessFormatter.java
+++ b/src/main/java/net/logstash/logback/LogstashAccessFormatter.java
@@ -13,6 +13,7 @@
  */
 package net.logstash.logback;
 
+import ch.qos.logback.classic.spi.ILoggingEvent;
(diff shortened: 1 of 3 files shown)


philsttr:
Thanks!!

PR #394, opened 2020-03-11
Replace `Charset.forName("UTF-8")` to `StandardCharsets.UTF_8`

ArthurGazizov:
Hi，

Is it necessary to replace `Charset.forName("UTF-8")` or some codes like `new String(encoded, "UTF-8")` to the constant variable StandardCharsets.UTF_8 which JDK 7 has provided


philsttr:
Thanks!  (Leftovers from when this library supported java 1.6)

PR #398, opened 2020-04-16
typesafe slf4j Logger wrapper that only allows StructuredArgument whe…

lldata:
I am migrating a codebase that uses slf4j Logging with lots of patterns to one that uses structured logging, so the fields will be seachable in our log database.

I really like the StructuredArgument methods that you provide, but I would like a way to enforce that way of logging, rather than just allowing a mixed approch. So I came up with this PR. It needs polish, but I am submitting it now to have a discussion if you think this would be valuable. 

Example code before PR:
```java
import static net.logstash.logback.argument.StructuredArguments.*;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class ClassicExample {
  private static final Logger logger = LoggerFactory.getLogger(ClassicExample.class);

  public void work(String arg) {
    // the way of the future :)
    logger.info("structured {}", kv("arg", arg));
    // I would prefer a compile error here to make code migration to structured logging easier
    logger.info("unstructured {}", arg);
  }
}
```

Example after PR:
```java
  import net.logstash.logback.Logger;

  import static net.logstash.logback.argument.StructuredArguments.kv;

  public class StructuredExample {
    // see https://github.com/qos-ch/slf4j/pull/167
    private static final Logger logger = Logger.getLogger();

    public void work(String arg) {
      // the way of the future :)
      logger.info("structured {}", kv("arg", arg));
      // compile error! You can still use slf4j Logger if you don't want this
      // logger.info("unstructured {}", arg);
    }
  }
```

So it is less boilerplate more type safety and you can still do it the old way if you like.

It would certainly provide value to projects that I am working on and have been working on in the past.

### DIFF
--- a/src/main/java/net/logstash/logback/Logger.java
+++ b/src/main/java/net/logstash/logback/Logger.java
@@ -1,0 +1,792 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.logstash.logback;
+
+import net.logstash.logback.argument.StructuredArgument;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Marker;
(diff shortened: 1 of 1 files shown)


philsttr:
Thank you for submitting a PR.  I really appreciate the thought, time, and effort you put into it.

I do not believe this is a good fit for logstash-logback-encoder.

It is not recommended to _only_ use StructuredArguments when logging.  Using "normal" arguments is a valid use case, and will be the most common use case used by applications.  Therefore, I do not think that introducing a `Logger` class that only allows StructuredArguments is very useful to the large majority of applications.

Additionally, I do not want to introduce a top-level Logger interface into logstash-logback-encoder.  The majority of applications should use the Logger interface from the primary logging library used by the application (e.g. slf4j)

PR #413, opened 2020-05-24
Add option to split log messages on newlines - fixes #404

### LINKED ISSUE #404: Handling multi-line log messages
When configuring handling of newlines in log messages, a useful addition would be the ability to output the `"message"` field as a JSON array, instead of a JSON string.

For example, with either of these logging statements:

```java
logger.error("first line{}second line", System.lineSeparator());
// or
logger.error("first line\nsecond line");
// or
logger.error("first line\r\nsecond line");
```

The output would be consistent:

```yaml
{
    "message": [
        "first line",
        "second line"
    ],
    ...
}
```

This would provide an unambiguous parsing format on the target system instead of relying on the source platform's newline representation.

metacubed:
This PR adds an option to output the `"message"` field as a JSON array, instead of a JSON string.

For example, with either of these logging statements:

```java
logger.error("first line{}second line", System.lineSeparator());
// or
logger.error("first line\nsecond line");
// or
logger.error("first line\r\nsecond line");
```

The output is consistent:

```yaml
{
    "message": [
        "first line",
        "second line"
    ],
    ...
}
```

This provides an unambiguous parsing format on the target system instead of relying on the source platform's newline representation.

### DIFF
--- a/src/main/java/net/logstash/logback/LogstashFormatter.java
+++ b/src/main/java/net/logstash/logback/LogstashFormatter.java
@@ -416,6 +416,53 @@ public class LogstashFormatter extends LoggingEventCompositeJsonFormatter {
         }
     }
     
+    /**
+     * Write the message as a JSON array by splitting the message text using the specified regex.
+     *
+     * @return The regex used to split the message text
+     */
+    public String getMessageSplitRegex() {
+        return messageProvider.getMessageSplitRegex();
+    }
+
+    /**
+     * Write the message as a JSON array by splitting the message text using the specified regex.
+     *
+     * <p>
+     * The allowed values are:
+     * <ul>
+     *     <li>Null/Empty : Disable message splitting. This is also the default behavior.</li>
+     *     <li>Any valid regex : Use the specified regex.</li>
+     *     <li><tt>SYSTEM</tt> : Use the system-default line separator.</li>
+     *     <li><tt>UNIX</tt> : Use <tt>\n</tt>.</li>
+     *     <li><tt>WINDOWS</tt> : Use <tt>\r\n</tt>.</li>
+     * </ul>
+     * </p>
+     * <p>
+     * For example, if this parameter is set to the regex {@code #+}, then the logging statement:
+     * <pre>
+     * log.info("First line##Second line###Third line")
+     * </pre>
+     * will produce:
+     * <pre>
+     * {
+     *     ...
+     *     "message": [
+     *         "First line",
+     *         "Second line",
+     *         "Third line"
+     *     ],
+     *     ...
+     * }
+     * </pre>
+     * </p>
+     *
+     * @param messageSplitRegex The regex used to split the message text
+     */
+    public void setMessageSplitRegex(String messageSplitRegex) {
+        messageProvider.setMessageSplitRegex(messageSplitRegex);
+    }
+
     public void addProvider(JsonProvider<ILoggingEvent> provider) {
         if (provider instanceof ArgumentsJsonProvider) {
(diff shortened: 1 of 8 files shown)


src/main/java/net/logstash/logback/LogstashFormatter.java
```diff
@@ -416,6 +416,22 @@ public void setEnableContextMap(boolean enableContextMap) {
...
+        messageProvider.setLineSeparator(lineSeparator);
+    }
+
+    public boolean isWriteMultiLineMessage() {
+        return messageProvider.isWriteMultiLineMessage();
+    }
+
+    public void setWriteMultiLineMessage(boolean writeMultiLineMessage) {
+        messageProvider.setWriteMultiLineMessage(writeMultiLineMessage);
+    }
```
philsttr, line 433: It's not clear that the "lineSeparator" is only used as a regex to split the message into an array if and only if "writeMultiLineMessage" is true.

Additionally, the term "writeMultiLineMessage" does not convey that the message field will be written as an array.  A message with embedded line separators could also be considered a "multi line message".

Perhaps, we can combine these two fields into a single String field named "messageSplitRegex".  When null/empty, the message will not be split, and will be written as a JSON string.  When non-null, the message will be split, and written as a JSON array.

src/main/java/net/logstash/logback/composite/loggingevent/MessageJsonProvider.java
```diff
@@ -27,18 +28,41 @@
     
     public static final String FIELD_MESSAGE = "message";
 
+    private Pattern lineSeparatorPattern = null;
+    private boolean writeMultiLineMessage = false;
```
philsttr, line 32: Same comment as LogstashFormatter regarding naming of these fields.  Combine into a single "messageSplitRegex" field.

Also add some javadoc for this field.

metacubed:
@philsttr, thanks for the review. I'll make the changes.

I can use the existing `SeparatorParser` class to handle the special strings `"SYSTEM"`, `"UNIX"`, `"WINDOWS"`, etc. Any other strings would be treated as raw regex patterns.

philsttr:
I just had another thought.  Do we really need regexes for splitting the message?  regexes can be computationally expensive.  I wonder if just searching for specific strings would be faster, maybe just using a StringTokenizer.

metacubed:
A couple of problems with `StringTokenizer`:

- It does not support multi-character delimiters. For example, specifying `\r\n` as the delimiter will split by `\r` and by `\n`, not the combination.

- It does not preserve consecutive delimiters. For example, `"Hello\n\nWorld"` will return `[ "Hello", "World" ]`, instead of the expected `[ "Hello", "", "World" ]`.

<br/>
Also note the Javadoc comment on `StringTokenizer`: 

> `StringTokenizer` is a legacy class that is retained for compatibility reasons although its use is discouraged in new code. It is recommended that anyone seeking this functionality use the `split` method of `String` or the `java.util.regex` package instead.

Both of those are regex-based.

philsttr:
Sounds good.  It's a shame there isn't an easy way to split a string that doesn't involve regexes.

metacubed:
The Apache Commons [StringTokenizer](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringTokenizer.html) might be worth exploring, but it adds a new dependency on `commons-text`.

philsttr:
Just use a regex.  I'd rather not introduce another dependency.

metacubed:
@philsttr, all right, looks like I've addressed all the review comments.

PR #424, opened 2020-09-09
ci: enable Java build matrix

sullis:
https://github.com/actions/setup-java


sullis:
WDYT? @philsttr

philsttr:
Thank you @sullis !

I think I'd rather limit this to LTS releases, plus the most recent interim release (rather than a bunch of interim releases).  But I'll make that change after I merge this.

PR #444, opened 2020-10-23
Close JsonGenerators after usage.

PascalSchumacher:
In general `JsonGenerator` should be closed, so that Jackson can recycle buffers etc..

It looks like `logstash-logback-encoder` does it own buffer management, so I am not sure if it is necessary.

Sorry for the noise if this is irrelevant.


philsttr:
LGTM!  Thank you!

Issue #456, opened 2020-12-28
Make DelegatingAsyncDisruptorAppender more resilient to exceptions thrown by child appenders


brenuart:
The `EventHandler` used by the `DelegatingAsyncDisruptorAppender` does the following:

1. Invoke `AppenderAttachable#appendLoopOnAppenders(event)`. This method loops through the attached appenders and calls `Appender#doAppend(event)` on each. If an appender throws an exception, the process is aborted and subsequent appenders are not called.
Is this a problem? I mean, wouldn't it be safer to try/catch exceptions and give a chance to all appenders to process the event? Looking at Logback's `AsyncAppenderBase` I noticed they behave the same and don't care about exceptions... Could be the rational is appenders are not supposed to throw exceptions?

2. Flush the output stream of attached `OutputStreamAppender`s. Same remark here: if an exception is thrown when flushing the first output stream, the remaining appenders won't be flushed at all.
The case may be slightly different from `doAppend()` in that appenders like the `OutputStreamAppender` won't throw an exception when they have IO issues when processing the event. They log an error status instead and proceed.
According to me, the `DelegatingAsyncDisruptorAppender` should do the same: wrap `OutputStream#flush()` calls within a try/catch and log an error status when an IOException is thrown.

What do you think?
If you agree with point (2), I can include the modifications in the PR I'm about to submit for #454.

philsttr:
Sounds good to me.

One thing to consider is that if an exception occurs while appending/flushing on one event, it seems there will be a high probability that the same exception will occur while appending/flushing the next event on the next call to `onEvent`.  This will result in an error status being logged pretty frequently.  Perhaps those error status messages should be throttled (for example, log the first one, and then only log a percentage of the subsequent errors until the next success)?  Or only logged on state changes?  What are your thoughts?

brenuart:
Indeed... The easiest is probably to log an error status the first time flush throws an exception and reset that state when it succeeds.
However, looking at `LogEventExceptionHandler` I would say this is true for any exception thrown by an `EventHandler`. May be we should find a more generic solution?

philsttr:
Hmm.  I'm thinking the `DelegatingAsyncDisruptorAppender` shouldn't make any assumptions about how the delegate appenders will behave if those delegate appenders throw exceptions.  As you discovered, `OutputStreamAppender` will swallow the exception and stop itself if it encounters an exception.    Other appenders might have different behavior, and be able to recover.

[...]

Issue #455, opened 2020-12-27
Possible NullPointerException in DelegatingAsyncDisruptorAppender when flushing appenders


brenuart:
The DelegatingAsyncDisruptorAppender may throw a NullPointerException at line https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/appender/DelegatingAsyncDisruptorAppender.java#L69 in the appender's output stream is `null`. 

This may appear very odd, but appenders like the RollingFileAppender may have a `null` output stream during rollover or even after if it failed to rotate the underlying file... I'm pretty sure the FileAppender will throw an exception before the DelegatingAsyncDisruptorAppender had a chance to flush at the end of the batch - so there is no point in checking for a null output stream. At least not until the FileAppender is "fixed" and can handle the null case gracefully...

philsttr:
> so there is no point in checking for a null output stream

So you're saying don't check for a null outputstream in DelegatingAsyncDisruptorAppender?  What would the fix be then?

brenuart:
That's not what I meant... I should have say "... _it **seems** there is no point in checking for a null output stream because it is likely an exception would be thrown earlier anyway_...

The `OutputStreamAppender#getOutputStream()` doesn't say anything about whether it may or not return a null output stream. IMHO it would be safer to check for null anyway.

My example around the RollingFileAppender is from experience. The output stream can be null for a very short time during rollover (fair enough) - but it can stay null for longer when it fails to roll. In this case the appender is usually broken and won't be able to recover by itself.

Sorry for not being clear the first time :-(

philsttr:
Closed by #457

Issue #454, opened 2020-12-27
LoggingEventAsyncDisruptorAppender should flush appenders implementing Flushable at end of a batch


brenuart:
Similar to #346, it would be nice if `LoggingEventAsyncDisruptorAppender`also flushes appenders implementing `java.io.Flushable` at the end of a batch.

I came across this issue because we have some custom appenders not extending OutputStreamAppender that would benefit from an explicit flush as well... What do you think ?

philsttr:
Seems reasonable to me.  Would you be willing to submit a PR?

philsttr:
Closed by #457

PR #457, opened 2020-12-29
Make DelegatingAsyncDisruptorAppender more resilient to exceptions + flush Flushable appenders at end of batch

brenuart:
This PR addresses the following issues:

- Wrap calls to appenders within a try/catch block and make sure exceptions thrown by one does not prevent other appenders from receiving the event (#456).
- Flush appenders at end of batch only if they are started (#456).
- OutputStreamAppender may occasionally return a null OutputStream (#455)
- Flush appenders implementing java.io.Flushable (#454)

### DIFF
--- a/src/main/java/net/logstash/logback/appender/DelegatingAsyncDisruptorAppender.java
+++ b/src/main/java/net/logstash/logback/appender/DelegatingAsyncDisruptorAppender.java
@@ -57,66 +57,66 @@ public abstract class DelegatingAsyncDisruptorAppender<Event extends DeferredPro
         /**
          * Whether exceptions should be reported with a error status or not.
          */
-    	private boolean silentError;
+        private boolean silentError;
         
         @Override
         public void onEvent(LogEvent<Event> logEvent, long sequence, boolean endOfBatch) throws Exception {
             
             boolean exceptionThrown = false;
-			for(Iterator<Appender<Event>> it=appenders.iteratorForAppenders(); it.hasNext(); ) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/logstash/logback/appender/DelegatingAsyncDisruptorAppender.java
```diff
@@ -49,29 +54,69 @@
     private final AppenderAttachableImpl<Event> appenders = new AppenderAttachableImpl<Event>();
     
     private class DelegatingEventHandler implements EventHandler<LogEvent<Event>> {
+        /**
+         * Whether exceptions should be reported with a error status or not.
+         */
+    	private boolean silentError;
```
philsttr, line 60: Looks like your editor is configured to use tabs.  Please convert your additions to use spaces to keep consistent with the rest of the file.

philsttr (changes requested):
Looks great!  Love the unit tests.

I just had one minor comment regarding spaces v tabs, but other than that it's perfect.

src/main/java/net/logstash/logback/appender/DelegatingAsyncDisruptorAppender.java
brenuart: Indeed..

brenuart:
Should I close the three related issues or do you prefer to do it yourself after adding the relevant entries in the change log ?

philsttr:
I just closed them.  Thanks for the PR!

Issue #461, opened 2020-12-30
Reduce GC pressure in CompositeJsonEncoder and *LogstashTcpSocketAppender


brenuart:
`CompositeJsonEncoder` implements the `Encoder` interface and therefore must return the encoded event as a byte array.
The implementation makes use of an intermediate ByteArrayOutputStream to collect the various parts produced by the formatter and prefix/suffix encoders. When done, the result is returned as a byte array.

A new ByteArrayOutputStream is initialised for every log event. It starts with an initial size of about 1Kb (+ prefix/suffix length) by default and grows if the formatter produces a larger output. If we are lucky and the initial size is large enough, this process allocates 2 byte arrays and 2 memory copy. If the buffer needs to grow, a new one is allocated (larger) and the content of the previous is copied into it. We end up with 3 allocations and 3 copy operations. 

This process is repeated for every event and imposes an extra overhead on the garbage collector.

Most of the time, the caller will write the output of the Encoder into an output stream. In this case, using an intermediate byte array isn't the most efficient design (well, I know, this is how Logback's Encoder interface is designed :-( ...

But maybe we could do better... I was thinking about introducing a new `StreamingEncoder` interface similar to this:

```
public interface StreamingEncoder {
  void encode(Event event, OutputStream stream) throws IOException;
}
```

`CompositeJsonEncoder` can be easily modified to implement this new interface alongside the existing `byte[] encode(event)` method. Then we can adapt `AbstractLogstashTcpSocketAppender` around lines [L598-L602](https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java#L598-L602) to tell the encoder to write directly into the output stream if it implements the new `StreamingEncoder` interface.

This would be highly efficient while preserving support for "legacy" encoders.

I made a first POC with this idea and everything looks OK.
What do you think ?
Do you see other areas/classes that could be optimised using a similar technique?

philsttr:
\<rant>
[...]

PR #472, opened 2021-01-05
Reduce memory allocations by writing directly into the output stream (#461)

brenuart:
Reduce GC pressure by writing directly into the output stream instead of returning a byte array.

Introduce a new (internal) StreamingEncoder interface to be implemented by Encoders that supports writing directly into the output stream instead of returning their results as a byte array.

Update both the AbstractLogstashTcpSocketAppender and the CompositeJsonEncoder to support this new interface. This should hoppefully reduce the amount of short-lived byte arrays created for each log event.

Review test case and make use of actual instances instead of _mocks_ (required to preserve state during test otherwise conditional operations like start/isStarted do not work as expected).

See #461 for more information.


brenuart:
Regarding commit https://github.com/logstash/logstash-logback-encoder/pull/472/commits/3d68f910e2133c5c7c6b268827d2e84b39ca279b, I'm not sure it is actually a good idea to make use of Jackson internal utilities like `BufferRecycler`, `ByteArrayBuilder` and `SegmentedStringBuilder`...

brenuart:
Additional notes:

I made another simpler version of this PR (not committed) with a custom ByteArrayOutputStream that grows by allocating additional chunks (to avoid the memory copy that occurs in the JDK version when it grows). Like Jackson's approach, the buffer recycles only the first chunk (whose size is determined by the `minBufferSize` configuration property) and discards the others. This works like a charm provided `minBufferSize` is set to an optimal value. The buffer is stored in a ThreadLocal with a SoftReference so it can be garbage collected should the vm decides to do so... 

For information, in production we ingest approximatively 15K events/s - for one hour during a working day, average JSON size is around `540 bytes`, p95 is at 602 and the largest is several kilobytes (holds a stack trace) . An initial buffer size of 1k or 2k seems to be a good default value.

I quite like this solution. I wonder if we need to go any further... After all, the rest of the chain is not optimised... look at the Layout for instance... it returns a String... and most Logback components are not designed to Stream their output...

philsttr:
In general, the StreamingEncoder implementation looks great!  I'm happy to incorporate it into logstash-logback-encoder.


> The release process returns the tail buffer to the recycler

This was surprising to me.  Thanks for the detailed analysis!  Good stuff.

A custom buffer recycler sounds perfectly reasonable, unless it is crazy complex.   Want to submit it as an alternate PR for comparison?

brenuart:
Hi Phil... I'm back with this optimisation. A few words about what I did so far...

The `ReusableByteBuffer` is some kind of fast alternative to JDK ByteArrayOutputStream with two notable differences. First off all it maintains multiple byte[] arrays when it needs to grow and avoids copying data as does the JDK implementation. Secondly, instead of discarding all the internal byte arrays when reset, the buffer always keeps the first. The idea is to set its size large enough to hold at least the "average" json event and reduce the number of required allocations.

`ReusableByteBuffers` acts as a _pool_ of `ReusableByteBuffer` instances. 
The first implementation used a ThreadLocal to keep one instance per thread. After some thought, it appears this strategy may lead to a higher-than-required memory consumption with as many buffers as we have threads accessing the Encoder. This may not be a problem with Async appenders (only one thread) but may become serious with other appenders like FileAppender or ConsoleAppender for instance. If you look carefully at their implementation you will notice that multiple threads can access the Encoder - only the writing into the underlying output stream is guarded.
The new implementation maintains a ConcurrentLinkedList with SoftReference to allocated buffers. A new buffer is created when the list is empty. They are discarded when the JVM is under high memory pressure and decides to reclaim them.

The `CompositeJsonEncoder` can be used as a "standard" Encoder returning a byte[] array or as a `StreamingEncoder` to render directly inside an outputStream passed as argument together with the event to render. As for now, only the LogstashTcpSocketAppender supports this method.
The encoder maintains its own pool of buffers. Their initial size is controlled by the existing `minBufferSize` parameter. These buffers are used by the legacy "encode" methods returning a byte array. When it needs to encode an event, it gets a buffer from the pool and asks the formatter to render it directly into the buffer. When done, the buffer is returned to the pool and reused by subsequent events.

The `CompositeJsonLayout` follows the same logic. As you noticed, it also uses the same reusable buffers as the CompositeJsonEncoder. Because it needs to return a String instead of bytes, it uses a Writer around the buffer and asks the json formatter to render the event using it. This strategy is similar to what was done before, except it is now moved inside the CompositeJsonLayout instead of in the CompositeJsonFormatter. This implementation is not as "highly" optimised as for the encoder... but AFAIK Layouts are somehow deprecated (correct?).

I made a small benchmark using JMH to compare this implementation with version 6.6. Here are the results:
- speed: +23%
- GC count: -68%
- GC Time: -64%
- Allocations: -76%

Not bad... especially with regards to the positive impacts on memory and GC.
But we can do even better! The bottleneck now is in the `CompositeJsonFormatter` that creates a new Jackson JsonGenerator for every event. I made some additional experiments by reusing the same JsonEncoder as long as the output stream does not change. This gives the following improvements compared to 6.6:
- speed: +46%
- GC count: -87%
- GC Time: -86%
- Allocations: -87%

This of course requires additional changes. I propose to cover it in a separate issue. The improvements provided by this PR already seem to be good enough ;-)

src/main/java/net/logstash/logback/util/ReusableByteBuffers.java
```diff
@@ -0,0 +1,99 @@
...
+        return new ReusableByteBuffer(initialSize);
+    }
+    
+    /**
+     * Get a buffer from the pool or create a new one if none is available.
+     * The buffer must be returned to the pool after usage by a call to {@link #releaseBuffer(ReusableByteBuffer)}.
+     * 
+     * @return a reusable byte buffer
+     */
+    public ReusableByteBuffer getBuffer() {
```
philsttr, line 72: I'd prefer this method to be named `acquire`, and the release method to be named `release`.

Reasoning: I think the "acquire/release" terminology is more common for object pools than "get/release".  And since this pool can only hold buffers, repeating "Buffer" in the method names is not needed.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
```diff
@@ -0,0 +1,240 @@
...
+    @Override
+    public void close() {
+        this.closed = true;
+    }
+
+
+    /**
+     * Return the number of bytes stored in this {@link ReusableByteBuffer}.
+     */
+    public int size() {
```
philsttr, line 146: The term "size" is currently used for two things... the number of bytes written to the buffer (here), and the initial size (in the constructor param and a field).

To differentiate the two, I'd suggest using the term "capacity" when referring to the allocated capacity of the buffer. And "size" when referring to the bytes actually written.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
```diff
@@ -0,0 +1,240 @@
...
+        this(INITIAL_SIZE);
+    }
+
+    /**
+     * Create a new {@link ReusableByteBuffer}
+     * with the specified initial capacity.
+     *
+     * @param initialBlockSize the initial buffer size in bytes
+     */
+    public ReusableByteBuffer(int initialBlockSize) {
```
philsttr, line 86: rename initialBlockSize -> initialCapacity

src/main/java/net/logstash/logback/util/ReusableByteBuffers.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.lang.ref.Reference;
+import java.lang.ref.SoftReference;
+import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
+
+/**
+ * A pool of {@link ReusableByteBuffer}.
+ * 
+ * <p>The pool is unbounded and can hold as many buffers as needed. Buffers are kept in the pool
```
philsttr, line 24: Unbounded is a little scary to me.

Under what conditions would this grow "too big" ?


Also, what are your thoughts on adding some monitoring capability for this?  That could be a future enhancement, but we could plan for it now.

philsttr (changes requested):
This looks really good!

The performance numbers you posted are impressive.

My comments are mostly minor.  Overall it's a really great implementation.

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
```diff
@@ -148,23 +133,19 @@ private JsonFactory createJsonFactory() {
...
-                .disable(JsonGenerator.Feature.FLUSH_PASSED_TO_STREAM);
+                .disable(JsonGenerator.Feature.FLUSH_PASSED_TO_STREAM)
 
+                /*
+                 * Don't let the json generator close the underlying outputStream and let the
+                 * encoder managed it.
+                 */
+                .disable(JsonGenerator.Feature.AUTO_CLOSE_TARGET);
+                
+                //FIXME these settings are mandatory - should we instead apply them *after* jsonGeneratorDecorator is applied (ie. in createGenerator()) ?
```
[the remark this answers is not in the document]
brenuart: Always mandatory. As you noticed, the two methods the formatter exposes to encode take both a stream or a writer for destination. We cannot let Jackson close them...
This is quite odd tho because the `JsonFactory#createGenerator(stream)` method clearly states that it won't own the stream and therefore won't close it. But since the AUTO_CLOSE_TARGET is enabled by default... the generator will close the stream when itself is closed.

I was thinking that since these two params are "absolutely" required, we'd better to apply them last on the JsonFactory after all other decorators are applied...

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @author brenuart
+ *
+ */
+public class ReusableByteBuffer extends OutputStream {
+
+    /**
+     * The default size of the initial buffer
+     */
+    public static final int INITIAL_SIZE = 1024;
```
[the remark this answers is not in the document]
brenuart: It could... it is currently only used by `ReusableByteBuffers` - so it can be made _package private_ for instance. However, I wonder if it would interesting for others to know the default size ?

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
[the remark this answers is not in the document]
philsttr: +1 makes sense to me.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
brenuart: Correct me if I'm wrong... what you propose is:
- keep the `size()` method as it is atm
- rename the `initialSize` constructor arg to `initialCapacity`

Is that correct?

philsttr: Yes.  I wrote this comment first in my review (since it's where I noticed size was being used for two things), before I mentioned size->capacity rename elsewhere.   Just wanted to explain my reasoning.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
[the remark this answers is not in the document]
philsttr: Let's reduce to package private for now.  We can always increase later if there is a solid use case for it.

src/main/java/net/logstash/logback/util/ReusableByteBuffers.java
brenuart: Yeah... say that if you have 200 concurrent threads requesting a buffer exactly at the same time, then you will have 200 buffers in the pool. But this is quite unlikely to happen... especially if the Encoder is relatively fast. You would have had exactly the same if buffers were kept in a ThreadLocal...
By using a ConcurrentLinkedQueue we limit the amount of buffers in the pool to what the maximum concurrency actually requires - which is not the case with a ThreadLocal.

As it is for the moment, entries will be removed from the pool only when the JVM needs some more memory...

If the Encoder is used by one of our async appender - then we know it will always be from the same thread and the pool size will always be one.
If the Encoder is used by other kinds of appender - then the pool can grow up to the maximum number of threads logging at exactly the same time - not more.

With this in mind, I'm not convinced we should think about alternative strategies to actively reduce/shrink the buffer other than the SoftReference... your opinion?

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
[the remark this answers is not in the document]
brenuart: What about `1024` bytes for the default capacity? Is it large enough? 
It is important to have a good default value here applicable to most use cases. May be we should also add a word about it in the documentation.. ?

src/main/java/net/logstash/logback/util/ReusableByteBuffers.java
philsttr: > If the Encoder is used by one of our async appender - then we know it will always be from the same thread and the pool size will always be one.
If the Encoder is used by other kinds of appender - then the pool can grow up to the maximum number of threads logging at exactly the same time - not more.

Got it.  Thanks for the explanation.  This makes me feel better about it being "unbounded".  Essentially, while it is technically true that it is unbounded... in practice it should never grow more than the number of threads accessing it concurrently.

I think this behavior is worth mentioning in the javadoc for the paranoid (like me) that get scared when they see "unbounded".  ;)

> With this in mind, I'm not convinced we should think about alternative strategies to actively reduce/shrink the buffer other than the SoftReference... your opinion?

I agree.  The current solution is good.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
brenuart: Hmmm...
`CompositeJsonLayout` already had a `setMinBufferSize` method that is now reused to set the _initial capacity_ of the buffers. To make things clear, the method should have the same name in `CompositeJsonEncoder`...

Knowing that, do you still want me to change the `initialSize` of the ReusableBuffer to `initialCapacity`? It is an internal implementation after all - and it doesn't have to follow the same naming as the "public" interfaces represented by the two composite layout/encoder...

philsttr: Sigh.  That's unfortunate. 

But yeah, let's use capacity when referring to allocated capacity in the new code.  I think it's much more clear when dealing with allocated vs used numbers.

src/main/java/net/logstash/logback/util/ReusableByteBuffer.java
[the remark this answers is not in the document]
philsttr: > What about 1024 bytes for the default capacity? Is it large enough?

Hmmm.  Good question.  I just took a quick glance over our logs, and the majority seem to fit within 1024.

Another question... Let's say this capacity is too low for one particular customer's usage... When they upgrade to this new version of logstash-logback-encoder, will their performance be _worse/comparable/better_ than the old behavior?

> It is important to have a good default value here applicable to most use cases. May be we should also add a word about it in the documentation.. ?

Yeah, probably worth a section in the readme.

> Are you aware of any strategy that would grow/shrink the buffer size dynamically ?

Not without increasing the complexity of this feature tremendously.  Let's use the simple solution first.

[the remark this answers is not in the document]
brenuart: > Let's say this capacity is too low for one particular customer's usage... When they upgrade to this new version of logstash-logback-encoder, will their performance be worse/comparable/better than the old behavior?

Performance will be similar to old behaviour but not worst. The old implementation relied on Jackson Recycler and related buffers. If you remember our previous discussion on the subject I highlighted the fact that we actually had close to no reuse of the allocated buffers at all. When doing my experiments, I compared the performance of the new buffering mechanism with _reuse disabled_ (buffers were never returned to the pool so a new one had to be created every time) against 6.6. Results were:
- Speed: -5% (slower than 6.6)
- GC count: -2%
- GC time: -1.5%
- Allocations: -1.5%

This is the worst case scenario. As you can see, performances are _similar_ to 6.6...

> Hmmm. Good question. I just took a quick glance over our logs, and the majority seem to fit within 1024.
`1024` was the default `minBufferSize` of `CompositeJsonEncoder`... I took the same default.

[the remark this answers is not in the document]
philsttr: Ok.  I'm fine with 1024, mainly just based on the logs that I looked at quickly.  I'd be ok doubling it too, but wouldn't want to go too much higher.  Your call.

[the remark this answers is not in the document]
brenuart: > Yeah, probably worth a section in the readme.
I added some words about it under "Composite Encoders/Layouts". Please review after I commit and don't hesitate to rephrase my English...

philsttr (approved):

philsttr:
Thanks again for your work on this!  Very cool improvement.

(And thanks for addressing all my code review comments.  I know I can be picky at times. ;)  Cheers!

brenuart:
👍
I noticed however that my PR changed the order of import statements :-( Should it be part of the checkstyle rules as well?

philsttr:
I think the import order was all over the place when I originally enabled the checkstyle rules.  I tried to create rules that matched the existing codebase as much as possible to minimize reformatting, and to "stop the bleeding".

Having said that, I'm not against defining import order rules and reorganizing the existing imports.  If there is a clear majority practice, I'd like to stick with it.  But I'm less concerned about changing imports than reformatting code in general because I rarely use git blame to look at imports.  I'm usually looking at actual code when using git blame, and therefore am much more hesitant to reformat the actual code.

PR #507, opened 2021-03-18
README maven logger should be runtime, not compile

paulvi:
should be runtime, not compile dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,14 @@ Maven style:
     <groupId>net.logstash.logback</groupId>
     <artifactId>logstash-logback-encoder</artifactId>
     <version>6.6</version>
+    <scope>runtime</scope>
 </dependency>
 <!-- Your project must also directly depend on either logback-classic or logback-access.  For example: -->
 <dependency>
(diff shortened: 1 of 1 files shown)


philsttr (changes requested):
Thanks for the contribution @paulvi !

I agree that the example should use runtime by default, since the most common usage of logstash-logback-encoder is xml-configuration only.

However, logstash-logback-encoder provides many java extension points, for example:
* [Custom Json Providers](https://github.com/logstash/logstash-logback-encoder#custom-json-provider)
* [Appender Listeners](https://github.com/logstash/logstash-logback-encoder#appender-listeners)
* [Json Factory and Generator Decorators](https://github.com/logstash/logstash-logback-encoder#customizing-json-factory-and-generator)

If any of these extension points are used, then compile scope is needed.

Therefore, if the default example in the readme is changed to runtime scope, then a note needs to be added that describes A) when runtime scope can be used safely, and B) when compile scope must be used instead.

paulvi:
Thanks,
README now has details

Issue #549, opened 2021-07-06
Bump license-maven-plugin to 4.1 (from 1.9.0)


brenuart:
The project is currently using `com.mycila.maven-license-plugin:maven-license-plugin:1.9.0` which is a pretty old version of the plugin. The plugin is now at version `4.1` and with a different GAV. The correct GAV is now:
```
   <groupId>com.mycila</groupId>
   <artifactId>license-maven-plugin</artifactId>
   <version>4.1</version>
```

The plugin configuration should also be added to the <pluginManagement> section of the POM so it also applies when the plugin is invoked from the command line. This would allow to check and/or format the source files without requiring a "full" build as follows:
```
mvn license:check
mvn license:format
```

IMHO the build should only verify if a valid license is present on source files and fail if not. In other words, the build should invoke the "check" goal instead of "format" as it is for the moment. When it fails, the user can update the files by invoking the plugin on the command line as follows:
```
mvn license:format
```
(same behaviour as with checkstyle for instance, except checkstyle doesn't provide a goal to "fix" issues automatically)

PR #550, opened 2021-07-06
Update to license-maven-plugin:4.1

brenuart:
Update POM to use the latest version of the Mycila license plugin (and correct GAV).
Move plugin configuration inside the <pluginManagement> section so the plugin can be manually invoked from the command line. Particularly handy to fix header issues automatically…
Change the build process to “check” the files only and fail the build if not valid.

Related issue: #549

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,8 +98,39 @@ Source Code Style
 -----------------
 
 Please follow the style used by the existing code in the repository.
-
 Rules are enforced by [checkstyle](src/checkstyle/checkstyle.xml).
+
+Java source files must include the following header at the top of the file, before the _package_ declaration:
+
+```
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+```
+
+The project makes use of the great https://github.com/mathieucarbou/license-maven-plugin[Mycila Maven License Plugin] to check for the presence of a valid header in source files during the build process.
+
+You can manually check your source files by invoking the plugin manually on the command line as follows:
+
+```
+mvn license:check
+```
+
+You can also ask the plugin to automatically update the header for you like this:
+
+```
+mvn license:format
+```
 
 
 License
@@ -107,3 +138,4 @@ License
 
 By contributing, you agree that the contributions will be licensed under the
 [Apache License 2.0](https://github.com/logstash/logstash-logback-encoder/blob/main/LICENSE).
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,7 +120,7 @@ Java source files must include the following header at the top of the file, befo
  */
 ```
 
-The project makes use of the great https://github.com/mathieucarbou/license-maven-plugin[Mycila Maven License Plugin] to check for the presence of a valid header in source files during the build process.
+The project makes use of the great [Mycila Maven License Plugin](https://github.com/mathieucarbou/license-maven-plugin) to check for the presence of a valid header in source files during the build process.
 
 You can manually check your source files by invoking the plugin manually on the command line as follows:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,7 +103,9 @@ Rules are enforced by [checkstyle](src/checkstyle/checkstyle.xml).
 Java source files must include the following header at the top of the file, before the _package_ declaration:
 
 ```
-/**
+/*
+ * Copyright 2013-${year} the original author or authors.
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
--- a/pom.xml
+++ b/pom.xml
@@ -9,6 +9,7 @@
     <name>Logstash Logback Encoder</name>
     <description>Provides logback encoders, layouts, and appenders to log in JSON and other formats supported by Jackson</description>
     <url>https://github.com/logstash/logstash-logback-encoder</url>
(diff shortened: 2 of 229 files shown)


brenuart:
I noticed that the license header is using the JAVADOC comment style:

```
/**   <---------
 * Licensed under the Apache License, Version 2.0 (the "License");
 * ...
 */
```

This means that a class may have TWO class-level javadoc sections: one with the license and a second with the actual class documentation. IMHO the license header should be enclosed in a _standard_ comment block like this:

```
/* <--------
 * Licensed under the Apache License, Version 2.0 (the "License");
 * ...
 */
```

If you agree with that, I can easily update the header in all source files to conform with this style as part of this PR...

philsttr:
Nice!

> If you agree with that, I can easily update the header in all source files to conform with this style as part of this PR...

Yes, please do.  I never noticed that.

philsttr:
Oh, while you're modifying the license stuff...  can you add a copyright to the header and each file to which it applied?  like this:

```
 * Copyright 2013-2021 the original author or authors.
```

See also #536 and https://github.com/spring-projects/spring-boot/issues/27021#issuecomment-865890138, and an [example](https://github.com/spring-projects/spring-boot/blob/515072b2e1001e70ef5542e77d4dee9ef3fadbfb/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L1-L15)

According to that comment, leave the template value in LICENSE as-is, but just apply the copyright to the header, and each file.

brenuart:
I can of course add the copyright in the header... but I'm a bit worried about the date range...
What should I use as end year? 
(1) Is it the "current year"? In this case, should we update all files next year?
(2) Is it the year of the last modification of the file? Then we need to configure the license plugin to get that information from Git... This is supported by the plugin but comes with many drawbacks. First performances are horrible. Second, you start having strange behaviour depending on whether the file is modified, staged, new or renamed. The rename is even more confusing as it depends on Git detecting the rename or not - and as you know this depends on ho similar the file is with another (and sometimes Git says the file is renamed although it is a brand new one)... We took that road at work for several years and finally decided to drop any "git" dates in the header!

philsttr:
There is no perfect solution, but this problem isn't unique to logstash-logback-encoder.

So, we could follow the same practice as others do, like spring-boot.

As part of this one-time task, stamp everything with 2021.

Just keep an eye out for any pull requests, and update the headers manually then.

If that becomes a problem, we can manually bump the year next year.  I'm not worried about a once-a-year maintenance task.

brenuart:
Oh... this odd - StackHasherTest#expected_hash_should_be_generated() is failing because of the changes in the license header??? It passes when I revert the header back to what it was before...

Here is part of the stacktrace:
```
org.opentest4j.AssertionFailedError: expected: <e30d4cae> but was: <48abcbb0>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
	at net.logstash.logback.stacktrace.StackHasherTest.expected_hash_should_be_generated(StackHasherTest.java:156)
```

Should I just change the expected value to `48abcbb0` to make it happy ?

brenuart:
Maybe we should remove the header check from the checkstyle rules as well?
Was:
```
    <module name="RegexpHeader">
        <property name="header" value="^/\*\*\n^ \* Licensed under the Apache License, Version 2\.0 \(the &quot;License&quot;\);"/>
    </module>
```

philsttr:
> StackHasherTest#expected_hash_should_be_generated() is failing because of the changes in the license header???

I'm guessing the hash changed because of line number changes (?)

> Should I just change the expected value to 48abcbb0 to make it happy ?

Yes

> Maybe we should remove the header check from the checkstyle rules as well?

I kind of like keeping the checkstyle rule, since checkstyle plugins in IDEs will highlight the problem (vs having to wait until a maven build)

brenuart:
> I kind of like keeping the checkstyle rule, since checkstyle plugins in IDEs will highlight the problem (vs having to wait until a maven build)
I see the point. 
Unfortunately, the rule is now failing since we changed the license header. Comment style is different (minor issue) - but first line is different as well. It now includes the "current year" - this means next year we'll have to change both the license template AND the checkstyle rule... Honestly I see little benefit in keeping that rule..

brenuart:
Build is still failing but because of the flaky `LogstashTcpSocketAppenderTest.testConnectOnPrimary:243` - this test fails from time to time - I should definitely have a look at it and make it more reliable... 
Could you please restart the build ?

philsttr:
Done.  And filed #560 to address flaky test

pom.xml
```diff
@@ -36,15 +37,14 @@
...
-        <lifecycle-mapping.version>1.0.0</lifecycle-mapping.version>
         <maven-bundle-plugin.version>5.1.2</maven-bundle-plugin.version>
         <maven-checkstyle-plugin.version>3.1.2</maven-checkstyle-plugin.version>
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-enforcer-plugin.version>1.4.1</maven-enforcer-plugin.version>
         <maven-gpg-plugin.version>3.0.1</maven-gpg-plugin.version>
         <maven-jar-plugin.version>3.2.0</maven-jar-plugin.version>
         <maven-javadoc-plugin.version>3.3.0</maven-javadoc-plugin.version>
-        <maven-license-plugin.version>1.9.0</maven-license-plugin.version>
+        <license-maven-plugin.version>4.1</license-maven-plugin.version>
```
philsttr, line 47: _nit_ put this in alpha order

CONTRIBUTING.md
```diff
@@ -98,12 +98,46 @@ Source Code Style
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+```
+
+The project makes use of the great https://github.com/mathieucarbou/license-maven-plugin[Mycila Maven License Plugin] to check for the presence of a valid header in source files during the build process.
```
philsttr, line 123: github doesn't render this link markup correctly

Issue #553, opened 2021-07-06
Maven Shade plugin complains about duplicate/overlapping classes/resources


brenuart:
The Maven Shade plugin complains about overlapping resources then building the uber jar:

```
[WARNING] disruptor-3.4.4.jar, logstash-logback-encoder-7.0-SNAPSHOT.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/
```

Fortunately the plugin seems to pick the right MANIFEST.MF - but nothing is said in the documentation about his resolution strategy. IMHO it would be safer to configure the plugin to simply exclude the MANIFEST.MF from shaded jar.

PR #554, opened 2021-07-06
Maven Shade: exclude MANIFEST and produce shaded SOURCE jar

brenuart:
- exclude MANIFEST from shaded jars to make sure we keep only the one from the project in the final result
- explicitly configure the plugin to <include> the Disruptor and <exclude> all the other dependencies (if not, new project dependencies will be shaded as well if the plugin configuration is not updated)
- produce shaded source as well (handy when debugging applications making use of the library)

Related issue: #553

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 .settings/
 .*.md.html
 
+dependency-reduced-pom.xml
+
 META-INF/
 target/
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,6 @@
 .settings/
 .*.md.html
 
-dependency-reduced-pom.xml
-
 META-INF/
 target/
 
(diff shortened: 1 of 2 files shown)


brenuart:
The plugin still issue some warnings after this change:

```
[INFO] Minimizing jar net.logstash.logback:logstash-logback-encoder:jar:7.0-SNAPSHOT
[WARNING] /Users/brenuart/Documents/projects/git/logstash-logback-encoder/target/classes (Is a directory)
[WARNING] /Users/brenuart/Documents/projects/git/logstash-logback-encoder/target/classes (Is a directory)
```

This is caused by the `minimizeJar` option. It seems to be a known issue (see https://github.com/apache/maven-shade-plugin/pull/83) and should be fixed in future releases (issue title is different but seems to share the same root cause).

PR #561, opened 2021-07-07
checkstyle rule for java import statements

brenuart:
- update checkstyle rule to assert proper ordering of java import statements
- update source files to conform with the decided ordering

### DIFF
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -36,7 +36,14 @@
         <module name="AvoidStarImport"/>
         <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
         <module name="RedundantImport"/>
-
+        <module name="ImportOrder">
+            <property name="groups" value="java,javax,net.logstash"/>
+            <property name="ordered" value="true"/>
+            <property name="separated" value="true"/>
+            <property name="option" value="above"/>
+            <property name="sortStaticImportsAlphabetically" value="true"/>
+        </module>
+        
         <!-- Checks for whitespace                             -->
         <!-- See https://checkstyle.org/config_whitespace.html -->
         <module name="EmptyForIteratorPad"/>
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -94,5 +94,6 @@
     </module>
     <module name="RegexpHeader">
         <property name="header" value="^/\*\n^ \* Copyright .* the original author or authors\.\n^ *\n^ \* Licensed under the Apache License, Version 2\.0 \(the &quot;License&quot;\);"/>
+        <property name="fileExtensions" value="java"/>
     </module>
 </module>
--- a/src/checkstyle/checkstyle.xml
+++ b/src/checkstyle/checkstyle.xml
@@ -36,11 +36,12 @@
         <module name="AvoidStarImport"/>
         <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
         <module name="RedundantImport"/>
+        <module name="UnusedImports"/>
         <module name="ImportOrder">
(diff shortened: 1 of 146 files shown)


brenuart:
(I squashed the commits locally then force push - may not be very wise tho... maybe I should let you squash them at the time you merge the PR?)

src/checkstyle/checkstyle.xml
```diff
@@ -36,7 +36,14 @@
         <module name="AvoidStarImport"/>
         <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
         <module name="RedundantImport"/>
-
+        <module name="ImportOrder">
+            <property name="groups" value="java,javax,net.logstash"/>
```
philsttr, line 40: remove javax

I think one group for java and javax without a separator between is sufficient.

philsttr:
> (I squashed the commits locally then force push - may not be very wise tho... maybe I should let you squash them at the time you merge the PR?)

I'm generally not a fan of force pushing _after_ comments have already been made on the review.   Seeing the diff of new commits help understand how the comments were resolved.

However, feel free to force push if there aren't any comments on the review yet.

src/checkstyle/checkstyle.xml
brenuart: Hmmm... I can have checkstyle putting both `java.*` and `javax.*` in the same group with a regular expression - but IDE like Eclipse doesn't like it :-(

The project isn't using any `javax.*` stuff anyway so, either:
- I remove the `javax` group and let them fall into the "other" category
- I keep it asis with possibly two "java" groups at the top - "just in case..." (project is not using any javax stuff anyway)

philsttr: Let's remove javax.

A lot of that namespace is slowly being replaced with jakarta anyway.

brenuart: I was wrong... we have a few `javax.net.ssl.*` stuff... removing the `javax` group gives something like this:

```
import java.io.BufferedOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.io.OutputStream;
import java.net.InetAddress;
import java.net.InetSocketAddress;
import java.net.Socket;
import java.net.SocketTimeoutException;
import java.net.UnknownHostException;
import java.util.ArrayList;
import java.util.Collections;
import java.util.Formatter;
import java.util.List;
import java.util.Optional;

import net.logstash.logback.appender.destination.DelegateDestinationConnectionStrategy;
import net.logstash.logback.appender.listener.TcpAppenderListener;
import net.logstash.logback.encoder.SeparatorParser;
import net.logstash.logback.encoder.StreamingEncoder;

import ch.qos.logback.core.net.ssl.ConfigurableSSLSocketFactory;
import ch.qos.logback.core.net.ssl.SSLConfigurableSocket;
import com.lmax.disruptor.EventHandler;
import com.lmax.disruptor.LifecycleAware;
import com.lmax.disruptor.RingBuffer;
import javax.net.SocketFactory;
import javax.net.ssl.SSLContext;
import javax.net.ssl.SSLSocket;
import javax.net.ssl.SSLSocketFactory;
```

Do you still like it? Personally I'm more in favour of a `javax` group at the top below `java`.. matter of taste ;-)
Up to you...

brenuart:
I also noticed that checkstyle is not enabled on _test_ sources...

I think we should enable it as well, at least to avoid issue with checkstyle integration in IDEs. I could not find a way to configure the Eclipse plugin to ignore test sources. This gives me quite a lot of warnings :(

src/checkstyle/checkstyle.xml
philsttr: Ok.  I was thinking it was prefix based, and that java would match java and javax.  But since that's not the case, we can keep it.

Issue #559, opened 2021-07-07
Allow async appenders to block when buffer is full instead of dropping event


brenuart:
Async appenders currently drop events when the queue is full. Instead, they are cases where it might be useful for the appender to *block* until some space is available. The idea is to use the async feature as a buffer to handle spikes of events without loosing anything if they can't all fit in the buffer. In this case the logging thread is blocked until some free space becomes available...

This usage pattern is a mix between async (event processing offloaded on a separate thread) and sync (not loosing anything).

philsttr:
Seems reasonable.

In addition to block-forever, I would also like to see block-with-timeout as an option.

brenuart:
What would be the use case? I mean, if the buffer is full it means you are logging faster than what the destination can cope with... To me, adding a timeout is like increasing the buffer: it gives you more chances to have the event enqueued and not lost. If the timeout expires, the event is lost just like if you had none. Well, I don't see any valid use case where it would help... Do you have some in mind ?

philsttr:
>  I mean, if the buffer is full it means you are logging faster than what the destination can cope with...

It could also mean that the application is having transient connection issues to the log destination.

> To me, adding a timeout is like increasing the buffer: it gives you more chances to have the event enqueued and not lost.

Yeah, it's like increasing the buffer, but without the memory usage side effect.

> If the timeout expires, the event is lost just like if you had none. Well, I don't see any valid use case where it would help... Do you have some in mind ?

What I was trying to avoid is blocking the application _forever_ if the connection to the downstream log destination fails and continues to fail (mainly thinking about the tcp appender).  But maybe that's what some applications need (?).  I personally do not want my production systems to lock up if the log destination has an issue.  But I guess there are some apps where locking up is desirable (maybe for batch processing or something not handling active requests).

I see your point regarding a timeout though.   And also if the buffer is consistently full (like when the connection to the log destination is failing), then if every log message has to wait for a timeout, then the app will just grind slowly, which is not much better than blocking forever.

Having said all that, a low timeout is roughly equivalent to just dropping immediately.  And a high timeout is roughly equivalent to blocking forever.  So, I'll drop the timeout as a requirement.  We can just offer two options, to block indefinitely or to drop immediately.

brenuart:
See #565 - I think we can support drop, block forever and block with timeout easily...

PR #565, opened 2021-07-08
Async append blocking with timeout

brenuart:
Please have a look at this first draft implementation for #559.

In short, the appender supports two distinct operation modes (set via `asyncMode`):
- `DROP` (the default): the appender drops events when the ring buffer is full. This is the legacy behaviour.
- `BLOCK`: the appending thread is blocked when the buffer is full until some free space becomes available. An optional timeout can be specified using the `retryTimeout` configuration property (disabled by default).

The RingBuffer does not allow me to "publish with a timeout" - it offers either a "try publish" or a "publish". The later blocks indefinitely until there is enough space in the buffer to accept the event. Even worst, waiting threads are not unblocked when the disruptor is shutdown.
Instead, I decided to tryPublish and sleep before retrying until the optional timeout expires. Sleeping is done using `LockSupport.parkNanos()` just like what the Disruptor does too. Retries are scheduled every 100ms which may appear to be too frequent (CPU?). The problem is there is no way I'm aware of that tells me when there is room in the ring buffer. May be I can hook into the EvenHandler and awake any waiting threads when an event has been processed... This means a queue of waiting threads and potential synchronisation issues and/or performance impacts...

Instead of `asyncMode` and `retryTimeout` (feel free to comment on the name of these properties) it may be more convenient to collapse them into a single `appendTimeout` configuration property whose semantic would be:
- `-1`: timeout is disabled, means block until space is available (aka the `BLOCK` mode)
- `0`: times out immediately (aka the `DROP` mode) - this would be the default behaviour
- `>0`: times out after the given time
When the appender times out, it drops the event and logs a warning as it does already.

@philsttr Feedback?

### DIFF
--- a/README.md
+++ b/README.md
@@ -625,8 +625,36 @@ For example:
   </appender>
 ```
 
-The async appenders will never block the logging thread.
+#### RingBuffer Full
+
+The async appenders will by default never block the logging thread.
 If the RingBuffer is full (e.g. due to slow network, etc), then events will be dropped.
+
+Alternatively, you can configure the appender to wait until space becomes available instead of dropping the events immediately. This may come in handy when you want to rely on the buffering and the async nature of the appender but don't want to loose any event in case of large logging bursts that exceed the size of the RingBuffer.
+
+The behaviour of the appender when the RingBuffer is controlled by the `appendTimeout` configuration property:
+
+| `appendTimeout` | Behaviour when RingBuffer is full                                      |
+|-----------------|------------------------------------------------------------------------|
+| `< 0`           | disable timeout and wait until space is available                      |
+| `0`             | no timeout, give up immediately and drop event (this is the *default*) |
+| `> 0`           | retry during the specified amount of time                              |
+
+
+Logging threads waiting for space in the RingBuffer wake up periodically at a frequency defined by  `appendRetryFrequency` (default `50ms`). You may increase this frequency for faster reaction time at the expense of higher CPU usage.
+
+When the appender drops an event, it emits a warning status message every `droppedWarnFrequency` consecutive dropped events. Another status message is emitted when the drop period is over and a first event is succesfully enqueued reporting the total number of events that were dropped.
+
+
+#### Graceful Shutdown
+
+When stopped, async appenders perform a _graceful shutdown_ waiting until all events in the buffer are processed and the buffer is empty.
+The maximum time to wait is configured by the `shutdownGracePeriod` parameter and is set to `1 minute` by default.
+
+Events still in the buffer after this period is elapsed are dropped and the appender is stopped.
+
+
+#### Wait Strategy
 
 By default, the [`BlockingWaitStrategy`](https://lmax-exchange.github.io/disruptor/docs/com/lmax/disruptor/BlockingWaitStrategy.html)
 is used by the worker thread spawned by this appender.
--- a/README.md
+++ b/README.md
@@ -648,9 +648,10 @@ When the appender drops an event, it emits a warning status message every `dropp
 
 #### Graceful Shutdown
 
-When stopped, async appenders perform a _graceful shutdown_ waiting until all events in the buffer are processed and the buffer is empty.
-The maximum time to wait is configured by the `shutdownGracePeriod` parameter and is set to `1 minute` by default.
+In order to guarantees that logged messages have had a chance to be processed by asynchronous appenders (including the TCP appender) and ensure background threads have been stopped, you'll need to [cleanly shut down logback](http://logback.qos.ch/manual/configuration.html#stopContext) when your application exits.
 
+When gracefully stopped, async appenders wait until all events in the buffer are processed and the buffer is empty.
+The maximum time to wait is configured by the `shutdownGracePeriod` parameter and is set to `1 minute` by default.
 Events still in the buffer after this period is elapsed are dropped and the appender is stopped.
 
 
@@ -771,7 +772,6 @@ e.g.<br/><tt>phasedBackoff{10,60,seconds,blocking}</tt></td>
 See [AsyncDisruptorAppender](/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java)
 for other configuration parameters (such as `ringBufferSize`, `producerType`, `threadNamePrefix`, `daemon`, and `droppedWarnFrequency`)
 
-In order to guarantees that logged messages have had a chance to be processed by asynchronous appenders (including the TCP appender) and ensure background threads have been stopped, you'll need to [cleanly shut down logback](http://logback.qos.ch/manual/configuration.html#stopContext) when your application exits.
 
 ### Appender Listeners
 
--- a/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
@@ -63,14 +63,11 @@ import com.lmax.disruptor.dsl.ProducerType;
  * for more information about the advantages of using a {@link RingBuffer} over a {@link BlockingQueue}.
  * <p>
  *
- * This appender will never block the logging thread, since it uses
- * {@link RingBuffer#tryPublishEvent(EventTranslatorOneArg, Object)}
- * to publish events (rather than {@link RingBuffer#publishEvent(EventTranslatorOneArg, Object)}).
- * <p>
- *
- * If the RingBuffer is full, and the event cannot be published,
- * the event will be dropped.  A warning message will be logged to
- * logback's context every {@link #droppedWarnFrequency} consecutive dropped events.
+ * The behavior of the appender when the RingBuffer is full and the event cannot be published
+ * is controlled by the {@link #appendTimeout} configuration parameter.
+ * By default the appender drops the event immediately, and emits a warning message every
+ * {@link #droppedWarnFrequency} consecutive dropped events.
+ * It can also be configured to wait until some space is available, with or without timeout.
  * <p>
  *
  * A single handler thread will be used to handle the actual handling of the event.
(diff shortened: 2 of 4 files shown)


brenuart:
Sorry for all these force pushes... `LogstachTcpAppenderTest#testReconnectOnOpen()` was failing on Github but could not reproduce it on my local machine. I had no other choice than experimenting directly on Github :-(

Anyway, this last commit (https://github.com/logstash/logstash-logback-encoder/pull/565/commits/64901f5acaeff1556f83ed7ac59bcc97f3b589a3) should fix https://github.com/logstash/logstash-logback-encoder/issues/569 as well.

brenuart:
(rebased on main)

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -422,57 +450,127 @@ public void stop() {
...
+         * Calling Disruptor#shutdown() will wait until all enqueued events are fully processed,
+         * but this waiting happens in a busy-spin. To avoid wasting CPU we wait for at most the configured
+         * grace period before asking the Disruptor for an immediate shutdown.
+         */
+        long deadline = getShutdownGracePeriod().getMilliseconds() < 0 ? Long.MAX_VALUE : System.currentTimeMillis() + getShutdownGracePeriod().getMilliseconds();
+        while (!isRingBufferEmpty() && (System.currentTimeMillis() < deadline)) {
+            LockSupport.parkNanos(SLEEP_TIME_DURING_SHUTDOWN);
+        }
+        
+        this.disruptor.halt();
```
philsttr, line 472: _nit_  extract into separate method

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -422,57 +450,127 @@ public void stop() {
...
+        
+        /*
+         * Shutdown executor service
+         */
         this.executorService.shutdown();
 
         try {
-            if (!this.executorService.awaitTermination(1, TimeUnit.MINUTES)) {
-                addWarn("Some queued events have not been logged due to requested shutdown");
-            }
```
philsttr, line 440: why remove the warning?

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -422,57 +450,127 @@ public void stop() {
...
+            // Wait before retry
+            //
+            long waitDuration = Math.min(this.appendRetryFrequency.getMilliseconds(), System.currentTimeMillis() - timeout);
+            if (waitDuration > 0) {
+                LockSupport.parkNanos(waitDuration * 1_000_000L);
+            }
+        }
+        
+        return true;
+    }
```
philsttr, line 573: This method could return true (incorrectly) if the appender is stopped while the loop is processing.

Also, I think the subtraction order is reversed in the waitDuration calculation on line 566.  I think it should be `timeout - System.currentTimeMillis()` since at this point currentTimeMillis < timeout (due to the check on line 560).  And, after correcting the order, the additional timeout check can be replaced by an else block on the waitDuration check.

So, addressing those comments, here's a possibility:

```java
    protected boolean enqueueEvent(Event event) {
        long deadline = this.appendTimeout.getMilliseconds() < 0 ? Long.MAX_VALUE : System.currentTimeMillis() + this.appendTimeout.getMilliseconds();

        while (isStarted()) {
            if (this.disruptor.getRingBuffer().tryPublishEvent(this.eventTranslator, event)) {
                return true;
            }
            // Wait before retry
            //
            long waitDuration = Math.min(this.appendRetryFrequency.getMilliseconds(), deadline - System.currentTimeMillis());
            if (waitDuration > 0) {
                LockSupport.parkNanos(TimeUnit.MILLISECONDS.toNanos(waitDuration));
            } else {
                break;
            }
        }

        return false;
    }
```

philsttr:
> Anyway, this last commit (64901f5) should fix #569 as well.

Did you mean #566 ?   I don't see how this PR addresses #569.  Did I miss something?

brenuart:
>>Anyway, this last commit (64901f5) should fix #569 as well.
>
>Did you mean #566 ? I don't see how this PR addresses #569. Did I miss something?

I actually meant #566. 
`stop()` now calls `disruptor.halt()` which shuts down the disruptor unconditionally and causes a call to `LifecycleAware#onShutdown` on event handlers implementing that interface - which is the case of the TcpEventHandler. The TCP connection is therefore closed whatever happens.

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: Because the executorService isn't actually processing any event, only housekeeping work. Events are processed on threads scheduled by the disruptor. Therefore I thought it was enough to issue a warning only if some events remained in the buffer at the end of the shutdown procedure.

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: > This method could return true (incorrectly) if the appender is stopped while the loop is processing.

Correct. And this was on purpose... 

First of all, events emitted by the application after the appender is stopped are silently discarded (a first warn is emitted tho). Loosing an event because the retry process is aborted during shutdown looks the same to me... so I was ok keeping it silent.
Now suppose the method return false when the appender is stopped as you propose, a warn status will be emitted saying an event is _dropped because the buffer is full_ (which is correct). I didn't feel comfortable with this and initially thought I had to return the actual cause to provide a more accurate cause... but that made the thing more complex. 

With al that said, I now realise that a _buffer full_ message is enough and we don't probably need anything more precise (what would be the added value?).

> I think the subtraction order is reversed in the waitDuration calculation on line 566
Correct... this shows that my test cases are a bit too optimistic :-(

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: I like to keep `start()` and `stop()` quite the same and the opposite of each other.
If I move the the code to stop the disruptor into a separate method, then I think we should do the same for the start... but I'm not sure it will add that much to the readability of the code.

philsttr: Ok, this just seemed like a nicely isolated piece of code that would be a good candidate to be extracted as a method.

It's essentially a re-implementation of disruptor.shutdown, which was previously called from stop.  So changing the stop method from calling this.disruptor.shutdown to this.shutdownDisruptor seemed rather nice.

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
philsttr: Got it.  Thanks

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
philsttr: Ok.  Thanks for the explanation.

I commented on this because the documented behavior...

> @return {@code true} when the even is successfully enqueued in the ring buffer

... is different than the implementation.   If the appender is shutdown while attempting to enqueue, the event won't be enqueued, but the method still returns true (contrary to the documented behavior).  Therefore, this seemed like a bug.

Also, when looking at the method in isolation (without looking at the calling code), returning true when the event is not enqueued seems... odd.   So the documentation seemed correct to me, and the implementation seemed incorrect.

Perhaps if the method detects a shutdown, the method could throw an exception, rather than returning true?  Then the caller could handle it by swallowing it (and not logging an event dropped warning), or logging a different warning.

I'm ok with whatever you decide, but regardless of your decision, please update the documentation to describe all cases.

philsttr:
> stop() now calls disruptor.halt() which shuts down the disruptor unconditionally and causes a call to LifecycleAware#onShutdown

Got it.  Thanks.

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: FYI I did not refactor the stop() method as you propose... It is likely to be affected by other up coming changes I have in mind ;-)

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: I moved back the "enqueue" logic inside the "append()" method as it was before - the method looks as it was before with an additional while loop.
Regarding the appender stopped when coming back from sleep, I decided to simply "warn" the user with the same message as what would be emitted if the application attempts to log juste after the appender is stopped.

brenuart:
The PR looks go enough to me now. 
@philsttr Tell me what you think...
I'd like to have it merged tho 'cause other changes I have in the pipe rely on it ;-)

README.md
```diff
@@ -625,9 +625,37 @@ For example:
...
+
+When the appender drops an event, it emits a warning status message every `droppedWarnFrequency` consecutive dropped events. Another status message is emitted when the drop period is over and a first event is succesfully enqueued reporting the total number of events that were dropped.
+
+
+#### Graceful Shutdown
+
+When stopped, async appenders perform a _graceful shutdown_ waiting until all events in the buffer are processed and the buffer is empty.
+The maximum time to wait is configured by the `shutdownGracePeriod` parameter and is set to `1 minute` by default.
+
+Events still in the buffer after this period is elapsed are dropped and the appender is stopped.
```
philsttr, line 654: It would be beneficial to mention how to gracefully stop logback here.

There's already a sentence about cleanly shutting down logback a bit further down in the Async Appenders section.  Could just move that sentence up to this sub-section.

> In order to guarantees that logged messages have had a chance to be processed by asynchronous appenders (including the TCP appender) and ensure background threads have been stopped, you'll need to [cleanly shut down logback](http://logback.qos.ch/manual/configuration.html#stopContext) when your application exits.

philsttr:
Looks good!  I'll squash and merge after you resolve the existing conflicts.

Issue #567, opened 2021-07-09
Creating a Disruptor instance with an Executor is deprecated in favour of a ThreadFactory


brenuart:
The `AsyncDisruptorAppender` creates Disruptor instances using the constructor that takes an `Executor` as argument.
This constructor has been deprecated in 2015 in favour of a `ThreadFactory`. The deprecation comments says:

> Use a {@link ThreadFactory} instead of an {@link Executor} as a ThreadFactory is able to report errors when it is unable to construct a thread to run a producer.

The deprecation has been introduced by commit https://github.com/LMAX-Exchange/disruptor/commit/42730561258d58e444cc051c04c5267bd7bcce7e and is in effect since version `3.3.4`.

The `AsyncDisruptorAppender` already maintains a `ThreadFactory`. I don't see any problem in passing it to the Disruptor instead of the executor service.

PR #570, opened 2021-07-14
Create Disruptor with a ThreadFactory instead of ExecutorService

brenuart:
See #567

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
@@ -36,7 +36,9 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import javax.net.SocketFactory;
@@ -274,6 +276,11 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
     private volatile CountDownLatch shutdownLatch;
 
     /**
+     * The {@link ScheduledExecutorService} used to execute house keeping tasks.
+     */
+    private ScheduledThreadPoolExecutor executorService;
+    
+    /**
      * Event handler responsible for performing the TCP transmission.
      */
     private class TcpSendingEventHandler implements EventHandler<LogEvent<Event>>, LifecycleAware {
@@ -443,7 +450,7 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                     }
                 } finally {
                     if (!Thread.currentThread().isInterrupted()) {
-                        getExecutorService().submit(() -> {
+                        executorService.submit(() -> {
                             /*
                              * https://github.com/logstash/logstash-logback-encoder/issues/341
                              *
@@ -805,7 +812,7 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                 }
                 long delay = TimeUnit.MILLISECONDS.toNanos(keepAliveDuration.getMilliseconds()) - (System.nanoTime() - basedOnNanoTime);
                 try {
-                    keepAliveFuture = getExecutorService().schedule(
+                    keepAliveFuture = executorService.schedule(
                         keepAliveRunnable,
                         delay,
                         TimeUnit.NANOSECONDS);
@@ -837,7 +844,7 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
                 }
                 long delay = writeTimeout.getMilliseconds();
                 try {
-                    writeTimeoutFuture = getExecutorService().scheduleWithFixedDelay(
+                    writeTimeoutFuture = executorService.scheduleWithFixedDelay(
                             writeTimeoutRunnable,
                             delay,
                             delay,
@@ -950,16 +957,14 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
         }
 
         if (errorCount == 0) {
-
+          
             encoder.setContext(getContext());
             if (!encoder.isStarted()) {
                 encoder.start();
             }
 
-            /*
-             * Increase the core size to handle the reader thread
-             */
-            int threadPoolCoreSize = getThreadPoolCoreSize() + 1;
+            int threadPoolCoreSize = 0;
+
             /*
              * Increase the core size to handle the keep alive thread
              */
@@ -972,7 +977,15 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
             if (isWriteTimeoutEnabled()) {
                 threadPoolCoreSize++;
             }
-            setThreadPoolCoreSize(threadPoolCoreSize);
+            this.executorService = new ScheduledThreadPoolExecutor(
+                    threadPoolCoreSize,
+                    getThreadFactory());
+
+            /*
+             * This ensures that cancelled tasks do not hold up shutdown.
+             */
+            this.executorService.setRemoveOnCancelPolicy(true);
+            
             this.shutdownLatch = new CountDownLatch(1);
             super.start();
         }
@@ -983,15 +996,29 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
         if (!isStarted()) {
             return;
         }
+        
+        super.stop();
+
         /*
          * Stop waiting to reconnect (if reconnect logic is currently waiting)
          */
         this.shutdownLatch.countDown();
-        super.stop();
+             
+        /*
+         * Stop executor service
+         */
+        this.executorService.shutdown();
+        try {
+            if (!this.executorService.awaitTermination(1, TimeUnit.MINUTES)) {
+                addWarn("Some queued events have not been logged due to requested shutdown");
+            }
+        } catch (InterruptedException e) {
+            addWarn("Some queued events have not been logged due to requested shutdown", e);
+        }
     }
 
     protected Future<?> scheduleReaderCallable(Callable<Void> readerCallable) {
-        return getExecutorService().submit(readerCallable);
+        return executorService.submit(readerCallable);
     }
 
     protected void fireEventSent(Socket socket, Event event, long durationInNanos) {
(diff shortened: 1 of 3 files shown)


philsttr:
After this change, the AsyncDisruptorAppender no longer uses its executorService.  The executorService is now only used by the AbstractLogstashTcpSocketAppender.

I think we have two options here:

1. Keep the executorService on AsyncDisruptorAppender for backwards compatibility in case there are other subclasses outside of logstash-logback-encoder using the executorService.  In this case, the javadocs should be updated to indicate its not used by AsyncDisruptorAppender directly, and kept for backwards compatibility for subclass usage only.
OR
2. Move the executorService down to AbstractLogstashTcpSocketAppender.

I'm leaning in favor of 1.  If you agree, can you update the javadoc of the field, and perhaps the getter as well?

We could also deprecate the executorService on AsyncDisruptorAppender with the intention of moving down to AbstractLogstashTcpSocketAppender later.  Thoughts?

brenuart:
I see your point. However, we are moving to `7.0` and removed deprecated methods which are likely to break other people's code (I know they were marked @deprecated for a while already but, you know, I doubt many moved away from them already ;-)

I'd like to start the 7 series with as less deprecation as possible... Removing the `ExecutorService` from `AsyncDisruptorAppender` without notice is indeed not very "friendly" but how many will be affected? The workaround is very easy - just create the executor service yourself from the `ThreadFactory` and you are back in business.

So I would go for option (2) + describe the workaround in the release note.
Do you think this is acceptable?

philsttr:
Yeah, I think 2 is acceptable since we're bumping the major version.

brenuart:
Ok - done.
I'll write a note about the removal of the ExecutorService in the draft release note after the PR is merged.

brenuart:
@philsttr Are you ok with these changes? Can I squash & merge?

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -950,16 +957,14 @@ public synchronized void start() {
...
             if (!encoder.isStarted()) {
                 encoder.start();
             }
 
-            /*
-             * Increase the core size to handle the reader thread
-             */
-            int threadPoolCoreSize = getThreadPoolCoreSize() + 1;
+            int threadPoolCoreSize = 0;
+
```
philsttr, line 967: Doesn't this need to start 1? (to account for the reader thread, which is always used).  I think that's why the original code incremented the core size by one here

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -645,21 +605,10 @@ protected void setEventTranslator(EventTranslatorOneArg<LogEvent<Event>, Event>
         this.eventTranslator = eventTranslator;
     }
 
-    protected ScheduledExecutorService getExecutorService() {
-        return executorService;
-    }
-
```
philsttr, line 651: probably should move this method down to AbstractLogstashTcpSocketAppender to keep source compatibility if anyone has custom subclasses

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -645,21 +605,10 @@ protected void setEventTranslator(EventTranslatorOneArg<LogEvent<Event>, Event>
...
     protected Disruptor<LogEvent<Event>> getDisruptor() {
         return disruptor;
     }
 
-    protected int getThreadPoolCoreSize() {
-        return threadPoolCoreSize;
-    }
-    protected void setThreadPoolCoreSize(int threadPoolCoreSize) {
-        this.threadPoolCoreSize = threadPoolCoreSize;
-    }
```
philsttr, line 661: Should move these down to AbstractLogstashTcpSocketAppender as well

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
brenuart: I thought about it but was wondering if it actually was a good idea. Using the executorService without proper coreSize sizing may result in unpredictable behaviour. Wouldn't be safer to not expose it and let people use their own ExecutorService should they need one ?

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
```diff
@@ -324,22 +346,12 @@ public void testReconnectToSecondaryOnWrite() throws Exception {
...
+        // First attempt at sending the event throws an exception while subsequent
+        // attempts succeed. This should force the appender to close the connection
         // and attempt to reconnect
-        doThrow(new RuntimeException())
-            .doReturn("event1".getBytes(StandardCharsets.UTF_8))
-            .when(encoder).encode(event1);
-        
+        doThrow(IOException.class)
+            .doNothing()
+            .when(outputStream).write(any(byte[].class), anyInt(), anyInt());
```
philsttr, line 354: It's unclear why this test needed to change for the refactoring of the executor service.  Can you clarify?

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
```diff
@@ -176,32 +176,51 @@ public void testReconnectOnOpen() throws Exception {
...
         
+        appender.start();
         verify(encoder).start();
         
-        doThrow(new RuntimeException()).doReturn("event1".getBytes(StandardCharsets.UTF_8)).when(encoder).encode(event1);
-        
         appender.append(event1);
```
philsttr, line 205: It's unclear why this test needed to change for the refactoring of the executor service.  Can you clarify?

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
brenuart: Correct. Don't know why I left it at zero... :-(

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
```diff
@@ -539,18 +556,10 @@ public void testReconnectToSecondaryOnKeepAlive() throws Exception {
...
-            .when(socket).connect(host("localhost", 10001), anyInt());
-
-        // Throw an exception the first time the the appender attempts to write in the output stream.
+        // Throw an exception the first time the the appender attempts to write in the output
+        // stream - this will be the keep alive message sent while on the primary destination.
         // This should cause the appender to initiate the reconnect sequence.
-        doThrow(SocketException.class)
+        doThrow(IOException.class)
             .doNothing()
             .when(outputStream).write(any(byte[].class), anyInt(), anyInt());
```
philsttr, line 564: It's unclear why this test needed to change for the refactoring of the executor service.  Can you clarify?

src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
philsttr: That is a excellent point.  I agree.  Let's keep it encapsulated and not provide access.

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
brenuart: Changing these tests was not actually required by the feature. 
The initial version hooks on the Encoder#encode() method to simulate the scenario. However, if the encoder is also a StreamingEncoder the streaming version of that method is invoked instead which makes the test brittle. In order to not make any assumption about the encoder (which is irrelevant in this test case) I decided to throw an exception from the output stream instead. This makes the test case more obvious (IMHO) and similar to the others already present in the same class...

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
brenuart: same reasons as above

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
brenuart: Not related to the feature covered by the PR (indeed). 
The initial version threw a SocketException. This may wrongly let you think that the test case is about connection related issues which is not the case. In fact, the test case is about any kind of IOException thrown when writing in the output stream. I decided to change the type of the exception thrown to suppress any ambiguity...

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
philsttr: Ok.  Thanks for clarifying. :)

Just want to make sure the same scenario is being tested and verified.  I was worried because the stubs for the connect method were removed.  But after a deeper look, I think the verifications ensure that the same scenario is tested.

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -950,16 +957,14 @@ public synchronized void start() {
...
                 encoder.start();
             }
 
-            /*
-             * Increase the core size to handle the reader thread
-             */
-            int threadPoolCoreSize = getThreadPoolCoreSize() + 1;
+            int threadPoolCoreSize = 1;
+
             /*
```
philsttr, line 968: This is an extreme nit, but please add the comment back about the reader thread . This will help future us understand why it is initialized 1.

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
brenuart: I should have made these modifications in a separate PR... Honestly, it took me some time to remember what these tests are doing. I wrote them a couple of years ago already and after this second read I realised that some were not so obvious and could be made more robust. I hope they will be easier to read next time...

src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
brenuart: You are right... I see what you mean.

brenuart:
I need to add some words in the release note about the removal of the ExecutorService from the AsyncDisruptorAppender. Could you please create a first draft of the release note? I haven't done this yet with Github... I'll take care to fill in the blanks afterwards.

philsttr:
Sure thing.  Give me a few mins.

brenuart:
Np. Time for me to go to sleep anyway ;-)

Issue #588, opened 2021-07-31
Fix javadoc issues and generate javadoc during normal builds


brenuart:
The Javadoc produced by the project has issues at many places. These issues are silently ignored because the plugin is configured with _doclet=none_.

Since the project is producing a javadoc artifact for others to consume, we should rather not ignore issues and fix them. Most javadoc issues are about missing or invalid argument names or invalid formatting constructs. Not saying that all methods must be fully documented, but at least the existing javadoc should be correct.

Also the javadoc should be generated during a _normal_ build to raise issues early before the project is released.

The plugin emits an error when a @param refers to an unknown method parameter - which is good. However, it raises a warning when @param is missing for some of the arguments. Proposition is to configure the plugin to fail on *warning* as well to capture those missing @param tags. 

This strategy is ok with JDK 1.8 but may become problematic on 1.15+ because recent versions of the javadoc tool chain also emit warnings when javadoc is missing on a public/protected method. This would require to fully document our code... 
We are running multiple builds on different JDK versions to test compatibility but official artefacts are produced from JDK 1.8 only. Proposition is to configure the plugin to ignore warnings on JDK 1.15+ until the code is fully documented.

An alternative approach is to build javadoc on 1.8 only and disable it in other builds. I would first try to build javadoc in every builds (with fail on warnings potentially disabled) to test and enforce javadoc compatibility as well...

PR #594, opened 2021-08-04
Fix javadoc issue and produce javadoc artifact during normal builds

brenuart:
- fix javadoc issues (errors and warnings) to be compliant with JDK16
- produce javadoc artifact during normal build (instead of waiting for a release build)
- fail on error and warnings
- disable fail on warnings on JDK 15+ until code is fully documented (warnings are emitted when a public/protected method is not documented)

Addresses issue #588.


pom.xml
```diff
@@ -301,6 +305,18 @@
...
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                        <phase>package</phase>
```
philsttr, line 316: _nit_ The default phase of the `jar` goal is `package`, so the phase does not need to be configured here.

PR #596, opened 2021-08-04
Prevent duplicate generation of sources and javadocs artefacts during a release build

### LINKED ISSUE #590: Duplicate "sources" artifact produced during a release build
Two "sources" artefacts are produced during a _release_ build.
A first one is created by the activation of the `release-profile` and a second by the `maven-shade-plugin`. Only the latter is created during a standard build.

Although the second overrides the first, both are attached to the build which causes a second attempt to upload/deploy the same artifact to the remote Maven repository. This is useless and may cause issues if the remote repository prevents release artifact from being overridden. 

The build procedure should be reviewed to keep only the artifact produced by the shade plugin.

brenuart:
closes #590

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -318,11 +318,6 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-release-plugin</artifactId>
-                <version>${maven-release-plugin.version}</version>
-            </plugin>
-            <plugin>
                 <groupId>com.mycila</groupId>
                 <artifactId>license-maven-plugin</artifactId>
                 <executions>
@@ -532,6 +527,16 @@
                         </defaultProperties>
                     </configuration>
                 </plugin>
+                
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-release-plugin</artifactId>
+                    <version>${maven-release-plugin.version}</version>
+                    <configuration>
+                        <!-- Use our own profile instead of the default "release-profile" -->
+                        <releaseProfiles>release</releaseProfiles>
+                    </configuration>
+                </plugin>
             </plugins>
         </pluginManagement>
     </build>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -580,6 +585,33 @@
...
+             of the legacy release-profile that is otherwise activated during a release 
+             build by the maven-release-plugin.
+             
+             Reasons are:
+             - the "sources" artifact is produced by the shade plugin. The one produced by
+               the legacy profile must be discarded.
+             - javadoc is already produced by normal builds.
+             
+             Note: maven-deploy-plugin:3.0.0 does not support the "updateReleaseInfo" configuration
+                   parameter anymore. This profile can be removed after upgrading.
```
philsttr, line 599: Hmm.  logstash-logback-encoder does not use the maven-deploy-plugin.  It uses the [nexus-staging-maven-plugin](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deployment) instead, which overrides the deploy goals.

So, I'm a little confused as to why the maven-deploy-plugin is configured here.

brenuart: The `release-profile` is inherited from Maven Super POM (see https://maven.apache.org/pom.html, Super POM, line 115). This profile is activated by the release plugin and has the Deploy Plugin in its <build> section.

Changes here are meant to remove the javadoc/sources stuff while keeping the rest "untouched".

Note that the "release-profile" will be removed in future Maven versions. Also, if you are 100% certain the deploy plugin is not used at all - then we can simply get rid of this profile and configure the release plugin to not active any at all (which will be simpler).

philsttr: > if you are 100% certain the deploy plugin is not used at all - then we can simply get rid of this profile and configure the release plugin to not active any at all (which will be simpler).

My understanding of the [nexus-staging-maven-plugin](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deployment) is that it replaces all the functionality of the maven-deploy-plugin...

From [its documentation](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deployment#ConfiguringYourProjectforDeployment-DeploymentwiththeNexusStagingMavenPlugin), it "automatically replaces the deploy goal invocation of the Maven Deploy plugin in the deploy Maven life cycle phase with the deploy goal invocation of the Nexus staging Maven plugin."


Was there a problem with the source/javadoc artifacts with previous logstash-logback-encoder versions?  Or is this change just an optimization to the build process?

brenuart: First of all I changed the build to produce/validate Javadoc during non-release builds as well. The idea was to detect issues _before_ the actual release.

Secondly, as described in #590, 2 sources artefacts were produced and attached to the build. Both were deployed with the second overriding the first one. This is not good and likely to cause issues at some point in the future for the following reasons:
- ordering is not "explicit" - I mean it is not clear in the pom which one should be kept. A small unrelated change may break it (it is fragile)
- deploying twice the same artifact (that is same GAV) may cause issue depending on how the repository is configured (allows redeploy or not) - not an issue atm tho
- there is long lasting issue in Maven around duplicates artefacts attached to the build (should both be kept or the last replaces the first, etc). AFAIK the current behaviour is likely to be changed in future Maven versions... At the end, the rule is: "don't rely on duplicates" ;-)

We are using the nexus-staging-maven-plugin as you mentioned. It looks like we could simply get rid of the release profile... I'll give it a try.

Issue #616, opened 2021-08-20
LogstashMarker#equals() and #hashCode() are unstable


brenuart:
Two LogstashMarker are equal if:
- they have the same name (immutable)
- they have the same _references_ (children markers)

Problem is additional markers can be added to the reference list after a LogstashMarker is created breaking the equals/hashCode methods. This can lead to unpredictable and strange behaviour when the marker is used in a Set or as a Key in a Map.

Ability to add children to a marker (references) is a standard Slf4J behaviour mandated by the marker interface. However, the legacy Slf4J BasicMarker implementation does *not* include the referenced elements in the equals/hashCode computation... only the LogstashMarker does. I'm not fully aware of all the Marker particularities and I can't see why the Logstash version should behave that way.

@philstr What's your opinion? Are you aware of the reasons? If none can be identified, may be we could just get rid of this behaviour and rely on the standard equals/hashCode... ?

philstr:
That sounds reasonable to me! But I think you tagged the wrong philstr@... :) - did you mean philsttr@?

brenuart:
My bad...  I meant @philsttr ...

philsttr:
I believe the [contains check](https://github.com/logstash/logstash-logback-encoder/blob/b949b90099ede6c7b927c0241c55b5336b6fbdbe/src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java#L92) in `add` is currently intended to check for logically equivalent markers.

However, I don't think the logical equivalence provides much value, so I'd be ok if the equals and hashcode were removed so that object identity comparison is used everywhere.

This might break some strange edge cases leading to "duplicate" markers being added, so will need to be added to release notes as a backwards incompatibility.

PR #617, opened 2021-08-20
Refresh LogstashBasicMarker implementation and remove synchronisation

### LINKED ISSUE #613: LogstashBasicMarker implementation is inefficient and out-off-sync with implementation provided by Slf4J
`LogstashBasicMarker` is a _now obsolete_ copy of org.slf4j.helpers.BasicMarker at version 1.7.12
The original implementation has been updated and improved since then. The version included in logback-logstash-encoder should be updated accordingly.

A few points:
- replace the use of Vector by a more efficient List implementation
- use a StringBuilder instead of a StringBuffer

Note: Slf4J BasicMarker has a package protected constructor and cannot be easily re-used in other project - reason why the project maintains its own copy.

brenuart:
fixes #613

### DIFF
--- a/src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
+++ b/src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
@@ -18,11 +18,11 @@ package net.logstash.logback.marker;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Vector;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.slf4j.Marker;
 
-/* Copy of {@link org.slf4j.helpers.BasicMarker} from slf4j-api v1.7.12, with minor changes:
+/* Copy of {@link org.slf4j.helpers.BasicMarker} from slf4j-api v1.7.31, with minor changes:
  * 1. make the constructor public so that it can be extended in other packages
  * 2. add getReferences() method
 
@@ -62,9 +62,13 @@ import org.slf4j.Marker;
 @SuppressWarnings("serial")
 public class LogstashBasicMarker implements Marker {
 
+    private static final String OPEN = "[ ";
+    private static final String CLOSE = " ]";
+    private static final String SEP = ", ";
+    
     private final String name;
-    private List<Marker> refereceList;
-
+    private final List<Marker> referenceList = new CopyOnWriteArrayList<>();
+    
     /*
      * BEGIN Modification in logstash-logback-encoder to make this constructor public
      */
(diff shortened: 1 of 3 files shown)


brenuart:
Also fixes #616

src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
```diff
@@ -62,9 +62,13 @@
...
 public class LogstashBasicMarker implements Marker {
 
+    private static final String OPEN = "[ ";
+    private static final String CLOSE = " ]";
+    private static final String SEP = ", ";
+    
     private final String name;
-    private List<Marker> refereceList;
-
+    private final List<Marker> referenceList = new CopyOnWriteArrayList<>();
```
philsttr, line 70: I realize this is coming from slf4j-api's implementation, but this seems like a step in the wrong direction, since it will result in a lot of unused lists being allocated.

If markers and/or structured args are used heavily by an application, this will have an impact on memory usage and garbage collection.

The previous logic of deferring list creation until a reference is added seems better from a memory and garbage collection standpoint.

Having said that, I don't have exact numbers to back these claims up, so probably better to go with slf4j-api's implementation until proven otherwise.

brenuart: I agree with you... creating the list lazily would be better. However, this makes supporting concurrent accesses without synchronisation far less easier...

AFAIK people are not supposed to create many markers - hence the MarkerFactory caching previously created markers. Is it still true with Logstash?

philsttr: LogstashBasicMarker is the superclass of all of the markers and structured argument implementations provided by logstash-logback-encoder.  Caching is not used for any of these.   So, for example, a marker (with an unused list) will be created every time this log line is executed.

```
logger.info("Foo {}", StructuredArguments.keyValue("bar", "baz"));
```

If that line is executed frequently, it'll create a lot of unused list garbage.

slf4j-api can cache the markers it creates, since those markers are simple string values, so they are easily cachable.

In any case, I understand the synchronization problems being addressed, so we can leave it as is.  It's prioritizing performance over memory usage.  Classic tradeoff.

brenuart: >  I understand the synchronization problems being addressed, so we can leave it as is. It's prioritizing performance over memory usage

But I'm also fighting against excessive memory allocations ;-)
The key point is whether LogstashBasicMarker must be thread-safe or not... IMHO, markers are not shared/mutated by multiple threads in most scenarios.

brenuart: @philsttr I'm tempted to "drop" the thread-safe nature of the marker... certainly if we consider what has been said in https://github.com/logstash/logstash-logback-encoder/issues/613#issuecomment-907665354

brenuart: @philsttr Back on the `referenceList` again... Several options:

(1) make it 'not thread-safe'. Start with a _null_ reference list and lazy initialise it with a standard ArrayList when needed. Behaviour becomes unexpected if the Marker is accessed from multiple threads concurrently. Note that a COW list won't probably make multi-threaded usage more predictable: a first call to _hasReferences()_ may return _true_ but the list may be empty when you start iterating over it if another thread removed the elements in the meantime.

(2) keep it "thread-safe" but use an AtomicReference to hold the list. The overhead when no reference is ever added to the marker is limited to this atomic reference.

(3) drop support for "child markers" (references) and make relevant methods "not implemented". Does it make sense to add additional markers to a LogstashMarker? 

What's your opinion?

philsttr: > BTW, what would be the use case? Can't we simply drop support for references and make the relevant methods "not implemented" ?

Unfortunately not, since the slf4j-api Logger methods only accept a single Marker argument.  Support for multiple markers is implemented by having markers reference others.


Regarding thread-safety.  I would rather keep it threadsafe.  Another option (3).... similar to (2) but instead of using an `AtomicReference`, we could use a volatile field accessed via a static `AtomicReferenceFieldUpdater`.   This prevents memory overhead of creating a lot of `AtomicReference` instances, at the expense of slightly slower access time and slightly more complex code.

PR #620, opened 2021-08-24
Implement exponential backoff for the retry strategy

### LINKED ISSUE #619: Bad Async performance under heavy load when appender configured to *not* drop events
The Async appenders show very bad performance when configured to not drop events when the ring buffer is full (`appendTimeout=-1`).
The async appender first _try_ publish the event in the ring buffer then sleeps for `appendRetryFrequency` milliseconds  before retrying. This relatively long sleep affects performance dramatically.

An alternative is to use an exponential backoff strategy capped by the `appendRetryFrequency`. This should give better performance while still limiting the CPU consumption.

brenuart:
The initial retries start very fast (a few nanoseconds pauses) and slow down up to the configure `appendRetryTimeout`.
In addition, limit retry to a single concurrent thread to preserve CPU in constraint environments.
These changes should give a better throughput with an acceptable latency when the queue is full.

Closes #619

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.LockSupport;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
```diff
@@ -259,13 +260,18 @@
      * Delay between consecutive attempts to append an event in the ring buffer when
      * full.
      */
-    private Duration appendRetryFrequency = Duration.buildByMilliseconds(50);
+    private Duration appendRetryFrequency = Duration.buildByMilliseconds(5);
```
philsttr, line 263: 5ms seems really fast.  

What is the reasoning behind this value?

PR #626, opened 2021-08-28
Adopt is no more, long live Zulu!

geertjanw:


philsttr:
Thanks for bringing this to our attention!

I've switched logstash-logback-encoder to Temurin instead of Zulu in #629

PR #631, opened 2021-09-06
Implement a pool of reusable JsonGenerator

### LINKED ISSUE #630: Reuse JsonGenerator instead of creating a new one for each Event
`CompositeJsonFormatter` creates a new Jackson `JsonGenerator` instance for every log event to serialise. This behaviour is inefficient and produces high pressure on the garbage collector under heavy load.

Instead, the formatter should attempt to reuse the JsonGenerators - possibly within a pool.

brenuart:
Implement a pool of reusable JsonGenerator similar to the ReusableByteBufferPool.
`CompositeJsonFormatter` now acts as a factory creating JsonFormatter instances bound to an OutputStream. The output stream is backed by a ReusableByteBuffer.
CompositeJsonEncoder/CompositeJsonLayout acquire a reusable JsonFormatter from the pool, write the event to it then flush the underlying buffer into a byte array, a string or an another output stream if it implements StreamingEncoder.
This approach reuses the JsonGenerator and the associated memory buffers. It also reduces the number of memory copy operations required to move data to destination.

fixes #630


brenuart:
Here are some benchmarks against 6.6 (test executed with `-Xmx64m` to reduce available memory, single threaded):

```
                6.6           7.0         Units
--------------------------------------------------------------------------
throughput      1.9320.000    3.713.000   ops/s    (+90%, nearly twice faster)
gc.count             1.739          230   counts   (-87%, 7.5 times less)
gc.time                644           90   ms       (-86%, 7.5 times less)
```

src/main/java/net/logstash/logback/layout/CompositeJsonLayout.java
```diff
@@ -82,26 +78,29 @@ public String doLayout(Event event) {
...
+        
+        } catch (IOException e) {
+            addWarn("Error formatting logging event", e);
+            return null;
+            
+        }
+    }
+
+    private void writeEvent(ReusableJsonFormatterPool<Event>.ReusableJsonFormatter cachedFormatter, Event event) throws IOException {
+        try (Writer writer = new OutputStreamWriter(cachedFormatter.getBuffer())) {
```
brenuart, line 93: Creating a new OutputStreamWriter every time is not very efficient... but was already the case before these changes. Layout support will be deprecated and removed in the next major release... therefore I decided not to spend too much time and complexity in a more efficient implementation.

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
```diff
@@ -118,6 +118,56 @@ public boolean isStarted() {
...
+        }
+        
+        public void writeEvent(Event event) throws IOException {
+            writeEventToGenerator(generator, event);
+        }
+        
+        public void dispose() throws IOException {
+            this.generator.close();
+        }
+    }
```
philsttr, line 168: _nit_ Perhaps rename to `close()` and make the class implement `java.io.Closeable`.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        
+        public void reset() {
+            this.buffer.reset();
+        }
+        
+        @Override
+        public void close() throws IOException {
+            release(this);
+        }
```
philsttr, line 81: It is unclear when users of this class should call dispose()/close()/reset().  Please document them to make it clear.

Also, can the visibility of dispose() and reset() be reduced?  I didn't see where they are used outside of the pool.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,140 @@
...
+            
+//        } else {
+//            // Encoder is stopped - dispose the JSON formatter instead of adding it
+//            // back into the pool
+//            try {
+//                cachedFormatter.dispose();
+//            } catch (IOException e) {
+//                // ignore
+//            }
+//        }
```
philsttr, line 137: Need to address the commented-out code.  If it is not needed, delete it.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        
+        public void write(Event event) throws IOException {
+            this.formatter.writeEvent(event);
+        }
+        
+        @Override
+        protected void finalize() throws Throwable {
+            dispose();
+        }
```
philsttr, line 67: Hmm.  Finalizers are not that great, since they are expensive for the gc to execute.  I tend to avoid them.  Is there an explicit need for it here?

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,140 @@
...
+            /*
+             * Create a new Deque if the SoftReference has been cleared.
+             * The new instance may be lost if another thread executes this code concurrently and the
+             * released JsonFormatter won't be reused. This behavior is expected and is far cheaper than
+             * implementing a more robust concurrent solution.
+             */
+            Deque<ReusableJsonFormatter> cachedFormatters = formatters.get();
+            if (cachedFormatters == null) {
+                cachedFormatters = new ConcurrentLinkedDeque<>();
+                formatters = new SoftReference<>(cachedFormatters);
```
philsttr, line 120: Is `formatters` referenced from multiple threads?  If so, need to make it `volatile`

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,140 @@
...
+
+import net.logstash.logback.composite.CompositeJsonFormatter;
+
+import ch.qos.logback.core.spi.DeferredProcessingAware;
+
+/**
+ * @author brenuart
+ *
+ */
+public class ReusableJsonFormatterPool<Event extends DeferredProcessingAware> {
```
philsttr, line 33: Would love some documentation on this.

What is it for?

Is it intended to be used outside of logstash-logback-encoder?

Is it threadsafe?

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
```diff
@@ -118,6 +118,56 @@ public boolean isStarted() {
...
+
+    /**
+     * Create a reusable {@link JsonFormatter} bound to the given {@link Writer}.
+     * 
+     * @param writer the {@link Writer} used by the {@link JsonFormatter}
+     * @return {@link JsonFormatter} writing JSON content in the writer
+     * @throws IOException thrown when unable to write in the writer or when Jackson fails to produce JSON content
+     */
+    public JsonFormatter createJsonFormatter(Writer writer) throws IOException {
```
philsttr, line 144: Is this method used anywhere?

philsttr (commented):
Can `ReusableByteBufferPool` be deleted?  (It looks like this review removes the only references to it, unless I missed something).

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: Indeed... I'll clean this up - was still a _wip_ to get your feedback ;-)

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: The underlying JsonGenerator needs to be explicitly disposed for Jackson to release the associated resources (return buffers to its pool). But honestly I don't know what are the consequences if it is not explicitly disposed and garbage collected... Jackson pools its memory buffers a bit the same way as we do here: a buffer is removed from the pool when used and added back when released. The pool does not maintain reference to the buffers that are currently in use - so chances are that the only consequence of not explicitly disposing the JsonGenerator would be not to return the buffer to the pool and have a _cache miss_ next time...

The "problem" here is we don't know when a JsonFormatter is not needed anymore and can be disposed - we rely on the SoftReference that holds the pool (Deque). JsonFormatter will be disposed when the GC decides it needs more memory - reason why I relied on the finalize method. IMHO the overhead should not be too high - and should occur only when running low in memory.

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
brenuart: It was initially used by the CompositeLayout. Not used since I decided to not spend anymore time optimising the Layouts... Can be removed.

brenuart:
> Can ReusableByteBufferPool be deleted? (It looks like this review removes the only references to it, unless I missed something).

Correct. It is not used anymore. Can be removed unless I find other places where it could be useful ;-)

For the background, you may have noticed that ReusableByteBufferPool uses a Deque of SoftReference whereas the ResuableJsonFormatterPool uses a SoftReference holding a Deque (the opposite). I changed the strategy because I noticed that when the GC decides to free the SoftReference it does it for nearly all of them in the same round. It is therefore more efficient to free the Deque instead of individual elements. At least I don't have to re-create a new SoftReference every time elements are added back to the pool... less garbage ;-)

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: Must be deleted.

brenuart:
About disposing JsonGenerator when a pooled JsonFormatter is GC... I had a look and the only consequence of not calling `JsonGenerator#close()` is that the internal memory buffers are not returned to Jackson's internal pool. As mentioned earlier this is not causing memory leak since Jackson does not seem to maintain any other reference to these allocated resources.

In fact, not closing the JsonGenerator is even better: memory is *not* returned to Jackson pools and is immediately reclaimed by the GC... which is exactly what we want. ReusableJsonFormatter are disposed by the GC when running low in memory - in this case it is better if the memory allocated by the JsonGenerator is freed as well.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: See https://github.com/logstash/logstash-logback-encoder/pull/631#issuecomment-914185865
Short version: implementation does not rely on `#finalize()` anymore.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: @philsttr Is wrapping the SoftReference in an AtomicReference better?

brenuart:
@philsttr I'm done with this PR. Everything looks good to me... Any additional comments?

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,223 @@
...
+        
+        /**
+         * Write the Event in JSON format into the enclosed buffer using the enclosed JsonFormatter.
+         * 
+         * @param event the event to write
+         * @throws IOException thrown when the JsonFormatter has problem to convert the Event into JSON format
+         */
+        public void write(Event event) throws IOException {
+            try {
+                this.formatter.writeEvent(event);
```
philsttr, line 121: What is the behavior if write is called after this ReusableJsonFormatter has been disposed of, or marked as not recyclable?  Should write throw an exception if this ReusableJsonFormatter has been disposed or if it has been marked as not recyclable before writing?

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,223 @@
...
+         * Dispose associated resources
+         */
+        protected void dispose() {
+            try {
+                this.formatter.close();
+                this.buffer.reset();
+                
+            } catch (IOException e) {
+                // ignore and proceed
+            }
```
philsttr, line 149: if formatter.close() throws, then buffer.reset() won't be called.  Probably should have two separate try/catch blocks, one for each, to ensure buffer.reset() is called even when formatter.close() throws.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
```diff
@@ -0,0 +1,223 @@
...
+     * <p>Note: usage is not thread-safe.
+     */
+    public class ReusableJsonFormatter implements Closeable {
+        private final ReusableByteBuffer buffer;
+        private final CompositeJsonFormatter<Event>.JsonFormatter formatter;
+        private boolean recyclable = true;
+        
+        ReusableJsonFormatter(ReusableByteBuffer buffer, CompositeJsonFormatter<Event>.JsonFormatter formatter) {
+            this.buffer = buffer;
+            this.formatter = formatter;
```
philsttr, line 101: _nit_ Use requireNonNull to make it clear that these fields can never be null

philsttr (commented):
Just had a few more minor comments.  Looks good otherwise!

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: You are right... However, `buffer.reset()` does nothing more than clearing the buffer content and dropping any extra buffers that may have been allocated. Since the ReusableJsonFormatter is "dropped" (no reference to it anymore) not calling reset() doesn't have any consequences.

Closing the JsonFormatter however makes it unusable -> the underlying JsonGenerator is closed and will throw an exception if used again.

Your remark makes me think that the implementation would be clearer if:
- dispose() close the JsonFormatter and nullify the references to the enclosed JsonFormatter and ReusableByteBuffer
- add a guard to the other method throwing an IllegalStateException if the ReusableJsonFormatter is re-used after being disposed

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
brenuart: > behaviour after it has been disposed

See https://github.com/logstash/logstash-logback-encoder/pull/631#discussion_r704724530

> behaviour when marked as not recyclable

JsonGenerator are stateful in the sense it knows for instance if the current structure of what has been generated so far allows for an object to be serialised. It would for instance complain if you attempt to `writeObject()` if a field name has not been output previously. There is unfortunately no way to "reset" it and revert to a clean initial state. The only solution I found is to avoid reusing it after it threw an exception. The "recyclable" field captures this information for when the ReusableJsonFormatter is released.
It can theoretically also throw an IOException if it has problem to write in the underlying OutputStream. This will never happen in our case.

Now, regarding your question... the current usage pattern is such that a ReusableJsonFormatter is used to write only one Event before it is released. But nothing prevents you from writing two events in a row with the same instance, they will be both written in the same buffer (that you can reset between the two write attempts). If you do so, I must say the outcome of the second write operation is unpredictable after the first threw an exception... It all depends on the JsonGenerator and what you are about to write.

I decided to keep the same behaviour has the JsonGenerator itself. The net result is similar to the following example when the first writeObjectField() operation throws an IOException:

```
JsonGenerator gen = ...
try {
  gen.writeObjectField("field1", obj1);
} catch(IOException e) {
  // ignore
}

gen.writeObjectField("field2", obj2);
```
... what would be the result? I don't know.. I haven't tested it... but such code doesn't look very wise :-(

Anyway, I can make the implementation more robust and, for instance, deny any further write() operation after an exception is thrown...

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
philsttr: >  Since the ReusableJsonFormatter is "dropped" (no reference to it anymore) not calling reset() doesn't have any consequences.

Ok.  The current code "looks" like a bug.   I've seen other cases like this that lead to sneaky hard-to-debug leaks.  So, seeing the cleanup statements lumped together in a single try block jumps out at me as a red flag.

If it is ok for reset() to not be called, then remove the call to reset() OR add a comment above it that says it's ok to not call reset() if close() throws an exception.

If reset() _should_ be called, then split the try blocks.

Doing any of the above will make it clear that the code is "correct", and not a potential problem.  

> Closing the JsonFormatter however makes it unusable -> the underlying JsonGenerator is closed and will throw an exception if used again.

Ok, that is enough to satisfy my concern.  Thanks for clarifying.   Adding more defensive code will make the error message more clear if the formatter was reused after dispose, but I don't think that is necessary since this class is meant to be internal to the library.

src/main/java/net/logstash/logback/util/ReusableJsonFormatterPool.java
philsttr: Ok.  Thanks for the explanation. :)

I think it's fine as-is.  If this were a user-facing class, I'd say we'd need to add some more defensive logic and better error messages.  But since this is internal to the library, we can keep it simple.

PR #634, opened 2021-09-06
Deprecate AbstractLogstashTcpSocketAppender#secondaryConnectionTTL

### LINKED ISSUE #633: Deprecate AbstractLogstashTcpSocketAppender#secondaryConnectionTTL
Property `secondaryConnectionTTL` belongs to the `PreferPrimaryDestinationConnectionStrategy` connection strategy and should not be exposed by the `AbstractLogstashTcpSocketAppender`.

This property has been kept in AbstractLogstashTcpSocketAppender for backwards compatibility purposes from before the connection strategies were implemented in #195 in 4.10. Previously (< 4.10) the concept of a connection strategy did not exist, and secondaryConnectionTTL was a just property on the appender. In 4.10, connection strategies were implemented, and the property remained on the appender for backwards compatibility.

As described in the documentation, the recommended way to configure this property is by setting it directly on the _preferPrimary_ connection strategy as illustrated in the following example:

```
      <connectionStrategy>
          <preferPrimary>
              <secondaryConnectionTTL>5 minutes</secondaryConnectionTTL>
          </preferPrimary>
      </connectionStrategy>
```

The property should be deprecated in a backward compatible way and marked for removal in the next major release (8.0).

(see https://github.com/logstash/logstash-logback-encoder/pull/576#issuecomment-907653080 for more background information)

brenuart:
Keep current behaviour for backward compatibility.
Emit a WARN status when the property is set.

fixes #633

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
@@ -1173,8 +1173,13 @@ public abstract class AbstractLogstashTcpSocketAppender<Event extends DeferredPr
      * @see PreferPrimaryDestinationConnectionStrategy#setSecondaryConnectionTTL(Duration)
      * @param secondaryConnectionTTL the TTL of a connection when connected to a secondary destination
      * @throws IllegalStateException if the {@link #connectionStrategy} is not a {@link PreferPrimaryDestinationConnectionStrategy}
+     * 
+     * @deprecated use {@link PreferPrimaryDestinationConnectionStrategy#setSecondaryConnectionTTL(Duration)} instead.
      */
+    @Deprecated
     public void setSecondaryConnectionTTL(Duration secondaryConnectionTTL) {
+        addWarn("Setting <secondaryConnectionTTL> on the appender is deprecated, set it on the connection strategy using <preferPrimary.secondaryConnectionTTL> instead.");
+
         if (connectionStrategy instanceof PreferPrimaryDestinationConnectionStrategy) {
             ((PreferPrimaryDestinationConnectionStrategy) connectionStrategy).setSecondaryConnectionTTL(secondaryConnectionTTL);
         } else {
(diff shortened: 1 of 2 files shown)


src/main/java/net/logstash/logback/appender/AbstractLogstashTcpSocketAppender.java
```diff
@@ -1173,15 +1173,30 @@ public Duration getReconnectionDelay() {
      * @see PreferPrimaryDestinationConnectionStrategy#setSecondaryConnectionTTL(Duration)
      * @param secondaryConnectionTTL the TTL of a connection when connected to a secondary destination
      * @throws IllegalStateException if the {@link #connectionStrategy} is not a {@link PreferPrimaryDestinationConnectionStrategy}
+     * 
+     * @deprecated use {@link PreferPrimaryDestinationConnectionStrategy#setSecondaryConnectionTTL(Duration)} instead.
      */
+    @Deprecated
     public void setSecondaryConnectionTTL(Duration secondaryConnectionTTL) {
+        addWarn("Setting <secondaryConnectionTTL> on the appender is deprecated, set it on the connection strategy using <preferPrimary.secondaryConnectionTTL> instead.");
```
philsttr, line 1181: `<preferPrimary.secondaryConnectionTTL>` is a little confusing, since it looks like xml, but is not the way the value would be set in xml.

Perhaps `<preferPrimary><secondaryConnectionTTL>` would be better?  I'm looking for better ideas. (?)

brenuart: I came across this issue several times already when throwing exceptions for invalid properties for instance.

I initially though that using an XML-like name was a good idea certainly if we consider that most users will use an XML file to configure their logging system. If we follow this approach then `<preferPrimary><secondaryConnectionTTL>` is certainly better.

Another option is to opt for a JavaBean-style format and do something like `"preferPrimary.secondaryConnectionTTL"`. But this looks weird to me...

In this particular case, maybe we could simply say `Setting <secondaryConnectionTTL> directly on the appender is deprecated. Instead you should set the connection strategy to <preferPrimary> and set <secondaryConnectionTTL> on it.`

What ever option we choose, we should probably review the other classes and standardise on the chosen naming convention...

philsttr: > `Setting <secondaryConnectionTTL> directly on the appender is deprecated. Instead you should explicitly set the connection strategy to <preferPrimary> and set its <secondaryConnectionTTL> property to the desired value.`

^ I like that.

> What ever option we choose, we should probably review the other classes and standardise on the chosen naming convention

Agreed.

PR #636, opened 2021-09-08
LogstashMarker refactoring

brenuart:
Most Logstash marker don't have any children/reference at all. Creating the referenceList upfront is a waste of memory and additional pressure on the GC. This PR modifies the implementation to create the reference list eagerly when the first reference is added. See https://github.com/logstash/logstash-logback-encoder/pull/617#discussion_r697888012 for the motivations behind this change.

### DIFF
--- a/src/main/java/net/logstash/logback/argument/DeferredStructuredArgument.java
+++ b/src/main/java/net/logstash/logback/argument/DeferredStructuredArgument.java
@@ -50,8 +50,15 @@ import org.slf4j.Logger;
  */
 public class DeferredStructuredArgument implements StructuredArgument {
 
+    /**
+     * Supplier for the deferred {@link StructuredArgument}
+     */
     private final Supplier<? extends StructuredArgument> structureArgumentSupplier;
 
+    /**
+     * Cached value of the structured argument returned by {@link #structureArgumentSupplier}.
+     * {@code null} until {@link #getSuppliedValue()} is first called.
+     */
     private volatile StructuredArgument suppliedValue;
 
     public DeferredStructuredArgument(Supplier<? extends StructuredArgument> structureArgumentSupplier) {
@@ -63,6 +70,12 @@ public class DeferredStructuredArgument implements StructuredArgument {
         getSuppliedValue().writeTo(generator);
     }
 
+    /**
+     * Get the deferred structure argument from the supplier or return it from the cache
+     * if already initialized.
+     * 
+     * @return the deferred structured argument
+     */
     private StructuredArgument getSuppliedValue() {
         if (suppliedValue == null) {
             synchronized (this) {
--- a/src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
+++ b/src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
@@ -26,7 +26,7 @@ import org.slf4j.Marker;
  * A simple implementation of the SLF4J {@link Marker} interface.
  */
 @SuppressWarnings("serial")
-public class LogstashBasicMarker implements Marker {
+class LogstashBasicMarker implements Marker {
 
     /**
      * The marker name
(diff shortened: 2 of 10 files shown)


brenuart:
In https://github.com/logstash/logstash-logback-encoder/pull/617#discussion_r697888012 we said we should probably reduce the visibility of `LogstashBasicMarker` to prevent people from using it as a base class for their own marker implementation (we don't want to support this use case).

You also mentioned that you kept LogstashBasicMarker as a separate class because it initially was a copy of SLF4J implementation. Having it in its own class was easier for maintenance/comparison with the original implementation.
This consideration does not apply anymore since "our" version differs slightly more from the original implementation (which has also changed in the meantime).

Considering these two points, I propose to merge `LogstashBasicMarker` inside `LogstashMarker` and keep a single base class for our all our marker implementations.

brenuart:
@philsttr `EmptyLogstashMarker` does not override equals/hascode. Equality is therefore solely on the marker name. Most (if not all) other logstash markers also require that "other" is of the exact same class. Shouldn't it be the case here as well ?

src/main/java/net/logstash/logback/marker/LogstashBasicMarker.java
```diff
@@ -107,6 +70,13 @@ public void add(Marker reference) {
...
         }
         
+        if (referenceList == null) {
+            synchronized (this) {
+                if (referenceList == null) {
+                    this.referenceList = new CopyOnWriteArrayList<>();
+                }
+            }
+        }
         referenceList.add(reference);
```
philsttr, line 80: _nit_ Referencing this field without using `this.`, and with using `this.`, in this single method is a little confusing when reviewing.  At first I thought they were different things.  Can you either remove `this.` everywhere, or add `this.` everywhere to use one approach.

src/main/java/net/logstash/logback/marker/MapEntriesAppendingMarker.java
```diff
@@ -99,13 +99,7 @@ public String toStringSelf() {
 
     @Override
     public boolean equals(Object obj) {
-        if (this == obj) {
-            return true;
-        }
```
philsttr, line 104: I think the == check should remain here (and in the other equals methods) as a short-circuit optimization to handle the case where the same object is being compared to itself.

src/main/java/net/logstash/logback/argument/DeferredStructuredArgument.java
```diff
@@ -50,8 +50,15 @@
  */
 public class DeferredStructuredArgument implements StructuredArgument {
 
+    /**
+     * Supplier for the deferred {@link StructuredArgument}
+     */
```
philsttr, line 55: 👍 Thanks for adding docs!

philsttr (commented):
> @philsttr EmptyLogstashMarker does not override equals/hascode. Equality is therefore solely on the marker name. Most (if not all) other logstash markers also require that "other" is of the exact same class. Shouldn't it be the case here as well ?

Yes, please add a check for the same class.

src/main/java/net/logstash/logback/marker/MapEntriesAppendingMarker.java
brenuart: This check is still there - but handled by `super.equals()`...
... or do you mean you'd prefer to keep the `==` to avoid an extra call to super() ?

philsttr: I'd prefer to keep the `==` check here to avoid executing `Objects.equals(this.map, other.map)` when the object is being compared to itself.

The `==` check in `super.equals(obj)` does not prevent the map comparison when the object is being compared to itself.

brenuart: Arghh.. yes indeed, you are right.. I missed that :-(
Let me revert these changes!

Issue #642, opened 2021-09-10
ObjectFieldsAppendingMarker may use wrong unwrapping JsonSerializer when multiple ObjectMapper are used


brenuart:
The `ObjectFieldsAppenderMarker` maintains a cache of "unwrapping" JsonSerializer to use to serialise the object value and avoid creating for each invocation. When no suitable JsonSerializer can be found in the cache, a new one is created. This new instance is created from the ObjectMapper returned by the JsonGenerator given as argument to the current invocation. It is then added to the cache with the object's class as key.

A second attempt at serialising a value of the same class will therefore use a JsonSerializer obtained from the cache. However, there is no guarantee that the JsonGenerator passed as argument uses the same ObjectMapper as the one used when creating the JsonSerializer obtained from the cache.

PR #643, opened 2021-09-10
Cache JsonSerializer "per" ObjectMapper

brenuart:
This PR adresses issue #642.

The primary objective of these changes is to maintain a separate cache of JsonProvider per ObjectMapper instance.

It also contains the following changes that are functional equivalent as the original implementation but delegate more of the work to Jackson itself:
- do not create a new instance of SerializerProvider ourselves but instead use `ObjectMapper#getSerializerProviderInstance` introduced in Jackson 2.7
- do not create a JsonSerializer for the class to serialize ourselves but instead delegate to Jackson `SerializerProvider#findValueSerializer(Class)` and therefore leverage its internal cache

### DIFF
--- a/src/main/java/net/logstash/logback/marker/ObjectFieldsAppendingMarker.java
+++ b/src/main/java/net/logstash/logback/marker/ObjectFieldsAppendingMarker.java
@@ -16,6 +16,7 @@
 package net.logstash.logback.marker;
 
 import java.io.IOException;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


brenuart:
As mentioned in the Javadoc, I'm also "worried" about maintaining a static cache of JsonProvider... :-(
There is an easy alternative however that does not require any caching at all. It basically relies on Jackson's ability to convert from one type to another... The `writeTo` method could be reduced to the following:

```
public void writeTo(JsonGenerator generator) throws IOException {
	if (this.object != null) {
		try {
			Map<String, Object> converted = mapper.convertValue(this.object, Map.class);
			for (Map.Entry<String, Object> entry: converted.entrySet()) {
				generator.writeObjectField(entry.getKey(), entry.getValue());
			}
		} catch (IllegalArgumentException e) {
			// thrown when object is not unwrappable
		}
	}
}
```

brenuart:
@philsttr The more I think about the alternative approach described above, the more I believe it is a very good alternative. Overhead seems to be limited to the intermediate map, without the burden of maintaining static caches... Seems to be a good tradeoff.

philsttr:
Creating an intermediary map every time the marker is written seems like too much overhead to me (for both memory and performance).

Using `mapper.convertValue` and then writing each entry from a Map is not guaranteed to result in the same JSON as the current behavior since it does not take some annotations like `@JsonSerialize` into consideration.  This will surprising to users expecting Jackson's behavior, with no clear way to revert to the old behavior.

The number of `ObjectMapper` instances will be small...  one per `CompositeJsonFormatter`, of which there are very few per application.

Given all the above, a static cache seems better.

philsttr (approved):

PR #646, opened 2021-09-14
Refactor JsonGenerator pooling

brenuart:
This PR is an extension of the work already done for issue #630.
Motivations:

(1)
For the pooling to be efficient, the feature `USE_THREAD_LOCAL_FOR_BUFFER_RECYCLING` must be disabled when creating JsonGenerator. If not, JsonGenerator creates additional buffers and re-use them per-thread. This pattern is not applicable in our case: there is no relationship between the JsonGenerator to use and the current thread.
Pooling Jsongenerator instances and the creation/configuration of these instances (disabling the feature) are therefore related and should ideally be implemented close together, in the same class.

(2)
Pooling is required only because CompositeJsonFormatter uses Jackson under the cover (JsonGenerator must be given an OutputStream when created). The pooling logic should therefore be isolated and hidden inside the CompositeJsonFormatter itself and considered an implementation detail. This would also lead to a cleaner interface with a single `write(Event event, OutputStream out)` method. This method can be used to write an event to whatever output stream without having to care about pooling at all...

(3)
The current implementation creates a ReusableByteBuffer and connects the JsonGenerator to it at creation time. They are both pooled at the same time. Content is first generated in the byte buffer before it can be copied in the output stream passed as argument to the write method. This intermediate buffer somehow limits the streaming capability of the implementation.
This commit now connects the JsonGenerator to a "DisconnectedOutputStream" when it is created. When the `write(event, out)` method is called, the output stream of the JsonGenerator is connected to the one passed as argument before the generator is invoked. Content produced by the generator is therefore written directly in the target output stream without requiring an intermediate buffer. It is now up to the caller to decide if it needs an intermediate buffer or not... Pooling and buffering are two separate concerns that are now handled separately.

(4)
The pooling logic is now handled by the `ObjectPool` class and is reused by both the ReusableByteBufferPool and the CompositeJsonFormatter.


src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
```diff
@@ -187,9 +263,6 @@ private JsonFactory decorateFactory(JsonFactory factory) {
     }
     
     protected void writeEventToGenerator(JsonGenerator generator, Event event) throws IOException {
-        if (!isStarted()) {
-            throw new IllegalStateException("Encoding attempted before starting.");
-        }
```
philsttr, line 192: Why is this removed?

philsttr (commented):
Looks good!  Just a couple minor comments.

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
brenuart: This method is called by `JsonFormatter#writeEvent` which in turn is called by `this#writeEvent` and that method is already guarded by the same condition.

src/main/java/net/logstash/logback/composite/CompositeJsonFormatter.java
```diff
@@ -119,40 +138,98 @@ public boolean isStarted() {
...
+            formatter.writeEvent(outputStream, event);
+        }
     }
     
     
-    public class JsonFormatter implements Closeable {
+    /**
+     * Create a reusable {@link JsonFormatter} bound to a {@link DisconnectedOutputStream}.
+     * 
+     * @param stream the output stream used by the {@link JsonFormatter}
```
philsttr, line 164: remove the @param   (sorry, it was in my previous suggestion, but I didn't explicitly call it out)

brenuart: Indeed... but weird... I expected the javadoc compiler to complain about this (it does if the @param is missing tho).

PR #655, opened 2021-09-21
Throw IllegalArgumentException at invalid time zone textual representation instead of defaulting to GMT

### LINKED ISSUE #650: FormattedTimestampJsonProvider#setTimeZone defaults to GMT if the zoneId is not a valid zone id
`FormattedTimestampJsonProvider#setTimeZone(ID)` delegates to `TimeZone.getTimeZone(ID)` to convert the supplied zone id into a valid ZoneId instance. However, as indicated by the javadoc, this method will not fail if the supplied zone id is not a valid zone identifier but returns GMT instead:

```
@return the specified <code>TimeZone</code>, or the GMT zone if the given ID cannot be understood.
```
 
This behaviour is confusing - FormattedTimestampJsonProvider should throw an IllegalArgumentException instead of silently revert to GMT in case of invalid zone id.

brenuart:
Closes #650.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1639,7 +1639,36 @@ You can change the timezone like this:
 </encoder>
 ```
 
-The value of the `timeZone` element can be any string accepted by java's  `TimeZone.getTimeZone(String id)` method.
+The value of the `timeZone` element can be any string accepted by java's `TimeZone.getTimeZone(String id)` method.
+
+It can be a valid time zone ID. For instance, the time zone ID for the U.S. Pacific Time zone is "America/Los_Angeles".
+
+If the time zone you want is not represented by one of the supported IDs, then a custom time zone ID can be specified to produce a TimeZone. The syntax of a custom time zone ID is:
+
+```
+   CustomID:
+      GMT Sign Hours : Minutes
+      GMT Sign Hours Minutes
+      GMT Sign Hours
+      
+   Sign: one of
+      + -
+   
+   Hours:
+      Digit
+         Digit Digit
+         
+      Minutes:
+         Digit Digit
+
+      Digit: one of
+         0 1 2 3 4 5 6 7 8 9
+```
+
+_Hours_ must be between 0 to 23 and _Minutes_ must be between 00 to 59. 
+For example, "GMT+10" and "GMT+0010" mean ten hours and ten minutes ahead of GMT, respectively.
+
+Use a blank string or `null` to use the default TimeZone of the system.
 
 
 ## Customizing LoggingEvent Message
--- a/README.md
+++ b/README.md
@@ -1640,35 +1640,8 @@ You can change the timezone like this:
 ```
 
 The value of the `timeZone` element can be any string accepted by java's `TimeZone.getTimeZone(String id)` method.
-
-It can be a valid time zone ID. For instance, the time zone ID for the U.S. Pacific Time zone is "America/Los_Angeles".
-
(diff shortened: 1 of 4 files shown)


src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
```diff
@@ -204,11 +204,60 @@ public void setPattern(String pattern) {
...
+     *         <i>Digit</i> <i>Digit</i>
+     * <i>Digit:</i> one of
+     *         <code>0 1 2 3 4 5 6 7 8 9</code>
+     * </pre></blockquote>
+     *
+     * <i>Hours</i> must be between 0 to 23 and <i>Minutes</i> must be
+     * between 00 to 59.  For example, "GMT+10" and "GMT+0010" mean ten
+     * hours and ten minutes ahead of GMT, respectively.
+     * 
+     * <p>Use a blank string or {@code null} to use the default TimeZone of the system.
```
brenuart, line 250: Joran does not allow to set a blank string nor a null value... May be we need another way here for the user to say it wants the default time zone of the system... ?

README.md
```diff
@@ -1639,7 +1639,36 @@ You can change the timezone like this:
...
+         
+      Minutes:
+         Digit Digit
+
+      Digit: one of
+         0 1 2 3 4 5 6 7 8 9
+```
+
+_Hours_ must be between 0 to 23 and _Minutes_ must be between 00 to 59. 
+For example, "GMT+10" and "GMT+0010" mean ten hours and ten minutes ahead of GMT, respectively.
```
philsttr, line 1669: I don't think logstash-logback-encoder should document the full format of the time zone string supported by java.util.Timezone, since logstash-logback-encoder is not responsible for parsing the string.   I don't want to have to keep this documentation up to date with the java.util.Timezone implementation/documentation.

I'd like to simplify this to something like...

The value of the `timeZone` element can be any string accepted by java's `TimeZone.getTimeZone(String id)` method. For example `America/Los_Angeles` or `GMT+10`

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
philsttr: Adding a special case for "DEFAULT" seems ok to me.

src/main/java/net/logstash/logback/util/TimeZoneUtils.java
```diff
@@ -0,0 +1,80 @@
...
+     * <i>Minutes:</i>
+     *         <i>Digit</i> <i>Digit</i>
+     * <i>Digit:</i> one of
+     *         <code>0 1 2 3 4 5 6 7 8 9</code>
+     * </pre></blockquote>
+     *
+     * <i>Hours</i> must be between 0 to 23 and <i>Minutes</i> must be
+     * between 00 to 59.  For example, "GMT+10" and "GMT+0010" mean ten
+     * hours and ten minutes ahead of GMT, respectively.
+     *
```
philsttr, line 58: Same comment as readme.  Let's not repeat the TimeZone string documentation here, but just give some examples and point to the TimeZone documentation for exact format.

src/main/java/net/logstash/logback/composite/FormattedTimestampJsonProvider.java
brenuart: `[DEFAULT]` can now be used to refer to the system default time zone (between brackets to follow same convention as used by the `pattern` property of the same class).

philsttr (approved):

PR #659, opened 2021-09-23
Allow FieldNames to override the field name of ContextJsonProvider when used in the context of IAccessEvent

### LINKED ISSUE #658: LogstashAccessEncoder#fieldNames does not change the field name used by ContextJsonProvider
Changing default field names on `LogstashAccessEncoder` by setting the `fieldNames` property does not change the name of the field used by the ContextJsonProvider.

The cause is the "instanceof" at line https://github.com/logstash/logstash-logback-encoder/blob/main/src/main/java/net/logstash/logback/composite/ContextJsonProvider.java#L57.

The "context" field is applicable to both IAccessEvent and ILoggingEvent and should therefore be declared in `CommonLogstashFieldNames` instead of `LogstashFieldNames`.

brenuart:
Closes #658

### DIFF
--- a/src/main/java/net/logstash/logback/composite/ContextJsonProvider.java
+++ b/src/main/java/net/logstash/logback/composite/ContextJsonProvider.java
@@ -18,7 +18,6 @@ package net.logstash.logback.composite;
 import java.io.IOException;
 
 import net.logstash.logback.fieldnames.LogstashCommonFieldNames;
-import net.logstash.logback.fieldnames.LogstashFieldNames;
 
 import ch.qos.logback.access.spi.IAccessEvent;
 import ch.qos.logback.classic.spi.ILoggingEvent;
@@ -53,9 +52,7 @@ public class ContextJsonProvider<Event extends DeferredProcessingAware> extends 
 
     @Override
     public void setFieldNames(LogstashCommonFieldNames fieldNames) {
-        if (fieldNames instanceof LogstashFieldNames) {
-            setFieldName(((LogstashFieldNames) fieldNames).getContext());
-        }
+        setFieldName(fieldNames.getContext());
     }
 
 }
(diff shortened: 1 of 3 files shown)


brenuart:
@philsttr Looks like the field `thread` could also be promoted to a _common_ name. The `ThreadNameJsonProvider` could then be used for both logging and access events.

Similarly, `SequenceJsonProvider` and `UuidProvider` are currently limited to ILoggingEvent but may potentially be useful in the context of IAccessEvent as well.

philsttr:
> @philsttr Looks like the field thread could also be promoted to a common name. The ThreadNameJsonProvider could then be used for both logging and access events.

The *FieldNames is only needed/used for the providers that are in LogstashAccessEncoder and LogstashEncoder by default.  *FieldNames is needed because the default providers of those encoders are not configured individually (like is done for the *CompositeJsonEncoder).  Since ThreadNameJsonProvider is not in LogstashAccessEncoder by default, it doesn't need to be in *FieldNames. 

For comparison, when the *CompositeJsonEncoders are used directly, *FieldNames is not used.  Instead, the field name used by each provider is configured directly on each provider, as opposed to being configured on the encoder level.

> Similarly, SequenceJsonProvider and UuidProvider are currently limited to ILoggingEvent but may potentially be useful in the context of IAccessEvent as well.

Yes, they can be refactored to handle IAccessEvent or ILoggingEvent, but they still wouldn't need to to be in *FieldNames, since they would not be in LogstashAccessEncoder by default.

philsttr (approved):

brenuart:
Ok - got it for Field Names.
Should we open separate issues to refactor `ThreadNameJsonProvider`, `SequenceJsonProvider` and `UuidProvider` to be usable for both ILoggingEvent and IAccessEvents ?
I'm not convinced many people will use them with IAccessEvents (should anybody use access events at all...) but this could make the code base consistent. It wasn't clear to me at first sight why these providers were limited to logging events... In fact, there is no reason except nobody asked for it ;-)

philsttr:
Yes, please open an issue for them. I agree we should make them consistent. Thanks :)

Issue #662, opened 2021-09-23
Allow UuidProvider to be used with IAccessEvent


brenuart:
`UuidProvider` is currently restricted to `ILoggingEvent` but may be useful with IAccessEvent as well.

brenuart:
`net.logstash.logback.composite.loggingevent.UuidProvider` is now deprecated and replaced by `net.logstash.logback.composite.UuidJsonProvider`. This new version is usable for both ILoggingEvent and IAccessEvent.

Issue #661, opened 2021-09-23
Allow SequenceJsonProvider to be used with IAccessEvent


brenuart:
`SequenceJsonProvider` is currently restricted to `ILoggingEvent` but may be useful with IAccessEvent as well.

brenuart:
`net.logstash.logback.composite.loggingevent.SequenceJsonProvider` is now deprecated in favour of the common `net.logstash.logback.composite.SequenceJsonProvider`. The later is usable for both ILoggingEvent and IAccessEvent.

Issue #660, opened 2021-09-23
Allow ThreadNameJsonProvider to be used with IAccessEvent


brenuart:
`ThreadNameJsonProvider` is currently restricted to `ILoggingEvent`but may be useful with IAccessEvent as well.

PR #663, opened 2021-09-23
Refactor some JsonProviders for use with both logging and access events

brenuart:
This PR addresses issues #660 , #661  and #662.


brenuart:
@philsttr Please have a first look at this PR when you have time...

The documentation lists the available providers in two sections: one with the providers applicable to ILoggingEvent and a second for IAccessEvent. I think it can be clearer if we move the providers applicable to _both_ in a third section that would appear before the two other.

philsttr:
> The documentation lists the available providers in two sections: one with the providers applicable to ILoggingEvent and a second for IAccessEvent. I think it can be clearer if we move the providers applicable to both in a third section that would appear before the two other.

Sounds good.  Make sure to link to the common section from both of the logging/access event specific sections

> Also the documentation says provider named message is available for access event... Isn't it rather accessMessage (cfr https://github.com/logstash/logstash-logback-encoder/blob/gh662-refactor-providers-for-logging-and-access/src/main/java/net/logstash/logback/composite/accessevent/AccessEventJsonProviders.java#L45-L47) ?

Yes, you are correct

philsttr:
Added the not-backward-compatible label since this change is not binary compatible (it looks like it should be source-compatible though)

philsttr (approved):

brenuart:
Conventions in the code base seems to prefix abstract classes with "Abstract" (I quite like it - this make it obvious it is an intermediate base class that cannot be used directly). 

To follow that convention, I renamed `FormattedTimestampJsonProvider` into `AbstractFormattedTimestampJsonProvider` but kept and deprecated the original for backward compatibility until next major version.

@philsttr Question: should I repeat the static fields in the now deprecated class (eg [AbstractFormattedTimestampJsonProvider.java#L40](https://github.com/logstash/logstash-logback-encoder/blob/gh662-refactor-providers-for-logging-and-access/src/main/java/net/logstash/logback/composite/AbstractFormattedTimestampJsonProvider.java#L40))? IMHO it is not strictly required...

brenuart:
> Also the documentation says provider named `message` is available for access event... Isn't it rather `accessMessage`...

`message` is probably a better name than `accessMessage`...:
- it better represents the actual purpose of the provider
- the default field name for this provider is `message`
- `LogstashFieldNames` also use `message`
- the documentation also refers to `message`

... only `AccessEventJsonProviders` uses a different name. Instead of changing the documentation, I would instead modify the `AccessEventJsonProviders` to align it with the rest.

README.md
```diff
@@ -1869,164 +1870,186 @@ These encoders/layouts make use of an internal buffer to hold the JSON output du
...
       </td>
     </tr>
+  </tbody>
+</table>
+
+
+### Providers for LoggingEvents
+
+The table below lists the available providers for _LoggingEvents_, and their configuration properties (defaults in parentheses).
+The provider name is the xml element name to use when configuring.
```
philsttr, line 1981: Mention and provide a link to the common providers section somewhere in this paragraph.

Perhaps:

```suggestion
The [common providers mentioned above](#providers-common-to-loggingevents-and-accessevents), and the providers listed in the table below, are available for _LoggingEvents_.
The provider name is the xml element name to use when configuring.  Each provider's configuration properties are shown, with default configuration values in parenthesis.
```

README.md
```diff
@@ -2048,208 +2071,173 @@ The provider name is the xml element name to use when configuring.
...
 </table>
 
 
 
-#### Providers for AccessEvents  
+### Providers for AccessEvents  
 
-The table below lists the available providers for AccessEvents, and their configuration properties (defaults in parentheses).
+The table below lists the available providers for _AccessEvents_, and their configuration properties (defaults in parentheses).
 The provider name is the xml element name to use when configuring.
```
philsttr, line 2156: Mention and provide a link to the common providers section somewhere in this paragraph.

Perhaps:

```suggestion
The [common providers mentioned above](#providers-common-to-loggingevents-and-accessevents), and the providers listed in the table below, are available for _AccessEvents_.
The provider name is the xml element name to use when configuring.  Each provider's configuration properties are shown, with default configuration values in parenthesis.
```

philsttr (commented):
> Question: should I repeat the static fields in the now deprecated class (eg AbstractFormattedTimestampJsonProvider.java#L40)? IMHO it is not strictly required...

To be binary-compatible, they would need to remain.  To be source-compatible, they can be removed.   I have tried to keep logstash-logback-encoder as-compatible-as-possible in the past, since that is best for its users.  But for 7.0 we've already broken compatibility a lot, so I'm ok with just keeping source-compatibility here.   We need to keep an eye on binary and source compatibility in future releases.  I don't want logstash-logback-encoder to get into the habit of breaking compatibility in lots of places all the time.

> CompositeJsonFormatter is another candidate... Should I rename it as well? And keep a deprecated version of it?

In a non-major release, I would say no (unless we kept full binary compatibility).  But you can rename it and keep source-compatibility only since 7.0 is a major release.

PR #665, opened 2021-09-26
Build with Java 17 LTS

### LINKED ISSUE #654: Build using Java 17 (LTS) instead of 16
The project should be built using LTS versions of Java. Currently these are 8, 11 and 17.
We should therefore build using 17 instead of 16.

brenuart:
Close #654 and #664

### DIFF
--- a/src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
+++ b/src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
@@ -25,7 +25,6 @@ import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
@@ -45,7 +44,6 @@ import java.net.SocketAddress;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.nio.charset.StandardCharsets;
-import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
@@ -108,8 +106,6 @@ public class LogstashTcpSocketAppenderTest {
     @Mock
     private TcpAppenderListener<ILoggingEvent> listener;
 
-    @Mock
-    private Random random;
     
     private class TestableLogstashTcpSocketAppender extends LogstashTcpSocketAppender {
         @Override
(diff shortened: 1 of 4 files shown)


brenuart:
Two comments about `RandomDestinationConnectionStrategy`...

(1)
The class used to expose a _public_ `getRandom()` method. The only reason these method was made public is to allow mocking by Mockito during test case (some tests are not in the same package so the method could not have been made protected instead).
Issue #664 is about Mockito unable to Mock the `java.util.Random` class with Java 17. To work around the issue, this PR does not expose the Random anymore but instead defines a `nextInt()` method directly on the strategy class. Test cases are modified to mock/intercept that method instead of mocking the Random itself.
However, this new method must still be _public_ which is definitely not "good". IMHO a better approach is to make it _protected_ and let testcases extend the class and override the method with the desired behaviour for the test.

(2)
`RandomDestinationConnectionStrategyTest` is about testing the _random_ strategy - but there is no "random" anymore: each test starts by giving the sequence of "random" number that should be returned by subsequent calls to `Random#nextInt()`... According to me, we are not testing the "RandomConnectionStrategy" here but rather the `DestinationConnectionStrategyWithTtl`...

philsttr:
(1) Feel free to make protected and extend as you mentioned.

(2) The test does not need to test that Random/ThreadLocalRandom behave properly, since they have been thoroughly tested by the jdk release process.  The test just needs to verify that the RandomDestinationConnectionStrategyTest calls Random/ThreadLocalRandom properly and returns their values.  So mocking the Random return values, and checking that the RandomDestinationConnectionStrategy returns those values is a valid test.

(3)  I'd have to dig in to see if it is safe to remove.  I don't think there's a harm in using a thread local Random to ensure the strategy is threadsafe.

src/test/java/net/logstash/logback/appender/LogstashTcpSocketAppenderTest.java
```diff
@@ -319,15 +315,15 @@ public void testRandomDestinationAndReconnectToSecondaryOnOpen() throws Exceptio
         appender.addDestination("localhost:10000");
         appender.addDestination("localhost:10001");
         RandomDestinationConnectionStrategy strategy = spy(new RandomDestinationConnectionStrategy());
-        doReturn(random).when(strategy).getRandom();
+        //doReturn(random).when(strategy).getRandom();
```
philsttr, line 318: remove commented out code

brenuart:
> (1) Feel free to make protected and extend as you mentioned.

I chose to pass a UnaryOperator function as argument to the constructor to generate random numbers. The default implementation makes use of `ThreadLocalRandom.current().nextInt()`. Test cases can provide their own implementation to control the sequence of generated random numbers.

> (3) I'd have to dig in to see if it is safe to remove. I don't think there's a harm in using a thread local Random to ensure the strategy is threadsafe.

I kept the `ThreadLocalRandom`. The "thread-safety" question can be raised for the other connection strategies as well. I propose to open a separate issue to cover them all if we feel the need to investigate any further.

PR #688, opened 2021-10-29
Enhance performance of AbstractPatternJsonProvider

brenuart:
This PR addresses multiple performance issues of the AbstractJsonPatternParser.

- PatternLayoutBase creates a new StringBuilder for every event it must serialize causing garbage and unnecessary pressure on the GC. This PR tries to reuse a StringBuilder per thread.
- Filtering of empty fields is retrieving values multiple times from the event. This may become become expensive when the value is computed by a PatternLayout. This PR uses Jackson's FilteringJsonGenerator instead.
- Make use of the most appropriate `JsonGenerator#write()` method instead of using `writeObject()` for "simple" value types. Using `writeObject()` delegates to the underlying ObjectCodec which is slower and produces lot more garbage.
- Refactor the code to make it easier to read and understand.

Benchmarks show that these changes make the AbstractPatternJsonProvider 4 times faster than LLE 6.6 and more importantly, it does not produce garbage anymore.


src/main/java/net/logstash/logback/composite/JsonWritingUtils.java
```diff
@@ -171,4 +178,91 @@ public static boolean shouldWriteField(String fieldName) {
...
+            generator.writeBoolean(((AtomicBoolean) value).get());
+            return;
+        }
+        if (value instanceof JsonNode) {
+            JsonNode node = (JsonNode) value;
+            
+            if (node instanceof NullNode) {
+                generator.writeNull();
+            } else {
+                node.serialize(generator, null); //FIXME
```
philsttr, line 262: FIXME?

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -259,323 +170,380 @@ protected Object transform(final String value) throws IOException {
...
-        return parseChildren(node);
     }
-
-    /**
-     * Return true if the given value is considered to be "empty".
-     * 
-     * @param value value to inspect
-     * @return {@code true} if the given value is considered to be "empty".
-     */
-    private boolean isEmptyValue(Object value) {
```
philsttr, line 546: The isEmptyValue method would result in filtering out empty collections, maps, and container nodes.  Does the new code still do that?  (I couldn't easily see where they were filtered out.  I only saw filtering out of null and empty strings)

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -259,323 +170,380 @@ protected Object transform(final String value) throws IOException {
...
+    protected static class LayoutValueGetter<Event> implements ValueGetter<String, Event> {
+        /**
+         * The PatternLayout from which the value is generated
+         */
+        private final PatternLayoutAdapter<Event> layout;
+
+        /**
+         * ThreadLocal reusable StringBuilder instances
+         */
+        private static final ThreadLocal<StringBuilder> stringBuilders = ThreadLocal.withInitial(StringBuilder::new);
```
philsttr, line 259: Rename to STRING_BUILDERS since this is static final to follow convention.

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -259,323 +170,380 @@ protected Object transform(final String value) throws IOException {
...
+        
+        /**
+         * StringBuilder whose length after use exceeds the maxRecylableSize will be
+         * discarded instead of recycled.
+         */
+        private static final int maxRecyclableSize = 2048;
```
philsttr, line 265: MAX_RECYCLABLE_SIZE

philsttr (commented):
Do the existing unit tests provide sufficient test coverage to ensure these changes did not change the generated JSON?

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
brenuart: Yes it does... otherwise the following test would fail: https://github.com/logstash/logstash-logback-encoder/blob/patternlayout-refactoring/src/test/java/net/logstash/logback/pattern/AbstractJsonPatternParserTest.java#L345 

> I couldn't easily see where they were filtered out.
Me neither... I did not check how the filtering mechanism works in detail but I notice empty collections were filtered out ;-)

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
brenuart: Ok with the case.
The idea here is to have a maximum length and avoid keeping very large string builders around (may be produced when rendering stack traces for instance). Considering that the string builder is used to render a single field of the pattern at once, `2048` is already a (very) large buffer... `1024` could be enough... what you think?

brenuart:
> Do the existing unit tests provide sufficient test coverage to ensure these changes did not change the generated JSON?

The seems to have good coverage already... Do you have additional scenarios in mind?

src/main/java/net/logstash/logback/composite/JsonWritingUtils.java
brenuart: Ooops... I forgot this one.

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
philsttr: 1024 seems good

philsttr:
> They seem to have good coverage already... Do you have additional scenarios in mind?

No.  I haven't looked at them.  I just wanted to confirm.  I trust your judgement.

src/main/java/net/logstash/logback/composite/JsonWritingUtils.java
brenuart: The FIXME was a reminder to add some comments with additional explanations about why the last argument is null. Comments are now added.

brenuart:
I'm done with this PR. 
@philsttr Are you ok with it?

philsttr (approved):

Issue #686, opened 2021-10-27
PatternJsonProvider: Inconsistent handling of configuration error


brenuart:
The behaviour of the PatternJsonProvider when it is given an invalid pattern differs depending on the type of error. Here are a few examples:

**(1) Invalid conversion word in Pattern layout**

The following JSON pattern refers to the log message using the `%message` conversion word instead of `%msg`:
```
{ "message: "%message" } 
```

The following error status is emitted when the provider is started but no exception is thrown:
```
09:58:13,556 |-ERROR in ch.qos.logback.core.pattern.parser.Compiler@4a87761d - There is no conversion class registered for conversion word [message]
09:58:13,556 |-ERROR in ch.qos.logback.core.pattern.parser.Compiler@4a87761d - [message] is not a valid conversion word
```

The provider is started and produces the following JSON output when serialising log events:
```
{"message":"%PARSER_ERROR[message]"}
```

**(2) Syntax error in Pattern Layout**

The following JSON pattern contains a syntax error (conversion word missing after `%`):
```
{ "message": "%" }
```

This is causing an ERROR status with an exception when the provider is started:
```
10:29:04,107 |-ERROR in ch.qos.logback.classic.PatternLayout("%") - Failed to parse pattern "%". ch.qos.logback.core.spi.ScanException: Unexpected end of pattern string
	at ch.qos.logback.core.spi.ScanException: Unexpected end of pattern string
	at 	at ch.qos.logback.core.pattern.parser.TokenStream.tokenize(TokenStream.java:119)
	at 	at ch.qos.logback.core.pattern.parser.Parser.<init>(Parser.java:68)
...
```

Unlike the previous case, the message field is now populated at runtime with an _empty string_ instead of a "%PARSE_ERROR" special value:
```
{"message": ""}
```

**(3) Invalid JSON pattern**

The PatternJsonProvider accepts a valid JSON string for its "pattern" property. If the supplied value is not a valid JSON, an ERROR status is logged and the provider defaults to an empty pattern, producing nothing at runtime.
Example (missing closing quote):
```
{ "message": "%message }
```

Produces the following ERROR when the provider is started:
```
10:32:55,479 |-ERROR in net.logstash.logback.composite.loggingevent.LoggingEventPatternJsonProvider@536aaa8d - [pattern] is not a valid JSON object
```

And the following JSON is produced at runtime:
```
{}
```

[...]

PR #691, opened 2021-11-03
Make AbstractJsonPatternParser throw an exception for invalid pattern instead of logging an ERROR status

brenuart:
`AbstractJsonPatternParser` now throws a `JsonPatternException` in case of invalid pattern. It is up to the `AbstractPatternLayoutJsonProvider` to catch the exception, log an error status and decide how it should behave.

Invalid PatternLayout formats are now detected and throw a `JsonPatternException` as well. Detection is done by  inspecting the parsed output returned by the PatternLayout when it is started, looking for the special %PARSER_ERROR marker inserted when it encounters an error.

The previous implementation used to ignore fields with an invalid pattern layout while keeping the other. This results in a partial JSON output with only the valid fields. This behaviour is now changed and the provider reverts to producing nothing in case of bad configuration.

These changes address issue #686.


src/main/java/net/logstash/logback/util/DelegatingContext.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class DelegatingContext implements Context {
+
+    private final Context delegate;
+    
+    public DelegatingContext(Context delegate) {
+        this.delegate = Objects.requireNonNull(delegate);
+    }
+
+    public StatusManager getStatusManager() {
```
philsttr, line 41: add `@Override` to appropriate methods from Context interface

src/main/java/net/logstash/logback/util/DelegatingContext.java
```diff
@@ -0,0 +1,100 @@
...
+import ch.qos.logback.core.Context;
+import ch.qos.logback.core.spi.LifeCycle;
+import ch.qos.logback.core.status.StatusManager;
+
+/**
+ * Logback {@link Context} implementation that delegate calls to another wrapped instance.
+ * 
+ * @author brenuart
+ */
+public class DelegatingContext implements Context {
```
philsttr, line 33: I'm a little worried about the Context interface evolving with new default methods, and us having to keep up with it.  (same for DelegatingStatusManager) Can you think of any other option that does not require these delegating objects

src/main/java/net/logstash/logback/util/DelegatingStatusManager.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class DelegatingStatusManager implements StatusManager {
+
+    private final StatusManager delegate;
+    
+    public DelegatingStatusManager(StatusManager delegate) {
+        this.delegate = delegate;
+    }
+
+    public void add(Status status) {
```
philsttr, line 37: add `@Override` to appropriate methods

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -85,76 +87,59 @@ public AbstractJsonPatternParser(final ContextAware contextAware, final JsonFact
...
-
+    protected abstract class TransformingOperation<T> implements Operation<Event, T> {
         @Override
-        public ValueGetter<Long, Event> createValueGetter(String data) {
-            return makeLayoutValueGetter(data).andThen(Long::parseLong);
+        public ValueGetter<Event, T> createValueGetter(String data) {
+            return makeLayoutValueGetter(data).andThen(this::convert);
         }
+        
+        protected abstract T convert(String value);
```
philsttr, line 112: _nit_ perhaps rename to `transform` so the method name relates to the class name

src/main/java/net/logstash/logback/pattern/AbstractJsonPatternParser.java
```diff
@@ -319,19 +303,26 @@ public String getValue(final Event event) {
...
+    private NodeWriter<Event> parseNode(JsonPointer location, JsonNode node) throws JsonPatternException {
         if (node.isTextual()) {
-            ValueGetter<?, Event> getter = makeComputableValueGetter(node.asText());
-            return new ValueWriter<>(getter);
+            try {
+                ValueGetter<Event, ?> getter = makeComputableValueGetter(node.asText());
+                return new ValueWriter<>(getter);
+            } catch (RuntimeException e) {
+                String msg = "Invalid JSON property '" + location + "' (was '" + node.asText() + "'): " + e.getMessage();
+                throw new JsonPatternException(msg, e.getCause());
```
philsttr, line 316: why `e.getCause()` and not `e`?

src/main/java/net/logstash/logback/util/DelegatingContext.java
brenuart: Errors detected by the PatternLayout fall in two categories:

- **Syntax Error**. Happens when the parser fails to parse the pattern for reasons like a missing closing bracket. In this case, the PatternLayoutBase logs an ERROR status and returns an empty string when its `doLayout(event)` method is invoked afterwards. Technically, the converter chain is empty (the _head_ is null).

- **Conversion Error**. Happens when an unknown conversion word is used (`%name`) or when it fails to instantiate an instance of the corresponding converter. In this case, instead of throwing an exception, the Parser replaces the conversion word with the string `%PARSER_ERROR[name]`.
Suppose you want to refer to the logger level and use `%loggerLevel` instead of `%level`, then the PatternLayout will produce  `"%PARSE_ERROR[loggerLevel]"` at runtime. From a technical point of view, the conversion chain contains a single LiteralConverter with a constant value.

We could rely on this behaviour and inspect the conversion chain looking for LiteralConverters with   `%PARSE_ERROR[...]` in them. This strategy has the following limitations:
- syntax errors are not detected. We can deduce a syntax error occurred via an other mechanism but we won't be able to capture the actual cause - only the fact a syntax error occurred;
- we know a conversion error occurred but here again we don't have the cause (the cause is logged in the error status);

Having said that, this may not be a problem: the PatternLayoutBase will have logged the actual cause itself before we detect an error occurred. The StatusManager would report something like this:
```
11:15:32,589 |-ERROR in ch.qos.logback.core.pattern.parser.Compiler@4944252c - There is no conversion class registered for conversion word [loggerLevel]
11:15:32,590 |-ERROR in ch.qos.logback.core.pattern.parser.Compiler@4944252c - [loggerLevel] is not a valid conversion word
11:15:32,591 |-ERROR in net.logstash.logback.composite.loggingevent.LoggingEventPatternJsonProvider@2781e022 - Invalid [pattern]: Invalid JSON property '/level' (was '%loggerLevel'): failed to interpret '%loggerLevel' (see previous statuses for more information)
```

The current solution emits a single ERROR status with all the details:
```
11:35:36,541 |-ERROR in net.logstash.logback.composite.loggingevent.LoggingEventPatternJsonProvider@2781e022 - Invalid [pattern]: Invalid JSON property '/level' (was '%loggerLevel'): There is no conversion class registered for conversion word [loggerLevel]
```

Both approaches can do the job.
I preferred to capture errors by hooking into `StatusManager#addError()` instead of parsing strings looking for error patterns. This looked more reliable to me... This must be balanced against the maintenance overhead you mentionned.

philsttr: As far as implementation... Both approaches are quiet "ugly" (no offense to your code, it's just what we're forced to do given logback's behavior).  But I think the approach from gh686-patternlayout-errorhandling-alternative is more maintainable.

I think in either case, the user will easily realize something is wrong.

So, I'm leaning toward the gh686-patternlayout-errorhandling-alternative impl

brenuart: Ok. Sold.

philsttr (approved):

PR #696, opened 2021-11-07
Allow masking of multiple substrings

### LINKED ISSUE #690: Only first ValueMasker is applied
**Describe the bug**
It's up for debate if this is a bug, but at least it is unexpected behaviour to me and I can't find it in the documentation. So filed a bug report anyway.

When configuring multiple `ValueMasker`s. and both of them would match a certain value, on the first one is applied.

**To Reproduce**
1. Assuming we have two `net.logstash.logback.mask.ValueMasker`s:
  * `com.example.MatchesOnStringA` which would mask all occurences of `"StringA"` to `"*******"`
  * `com.example.MatchesOnStringB` which would mask all occurences of `"StringB"` to `"*******"`
2. Use this `logback.xml` configuration†: 
  ```xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>

    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
            <jsonGeneratorDecorator class="net.logstash.logback.mask.MaskingJsonGeneratorDecorator">
                <valueMasker class="com.example.MatchesOnStringA"/>
                <valueMasker class="com.example.MatchesOnStringB"/>
            </jsonGeneratorDecorator>
        </encoder>
    </appender>

    <root level="INFO">
        <appender-ref ref="CONSOLE"/>
    </root>
</configuration>

  ```
  † this is maybe not entirely correct configuration, I've manually trimmed a few lines here and there to make it consise
  
3. Execute this logger statement:
  ```java
  log.info("StringA and StringB");
  ```
4. See the log 

**Expected behavior**
I would expect this to match 
```
******* and *******
```

But in fact only the first configured `ValueMasker` is applied, so the result is 
```
******* and StringB
```

**Additional context**
I had a look through the sources and I guess it's to be expected because [of this code](https://github.com/logstash/logstash-logback-encoder/blob/3e194210b7c3e6a11be2e8ca82b15986cd90d585/src/main/java/net/logstash/logback/mask/MaskingJsonGenerator.java#L527-L541):

```java
    private Object getMaskedValueForCurrentPathAndValue(Object value) {
        JsonStreamContext context = getOutputContext();
        for (ValueMasker valueMasker : valueMaskers) {
            Object maskedValue = valueMasker.mask(context, value);
            if (maskedValue != null) {
                return maskedValue;
            }
        }
        return null;
    }
```

philsttr:
Previously, masking of a single value would stop after the first ValueMasker returned a masked value.
This meant that only one ValueMasker would process a string.
Now, all ValueMaskers are given the opportunity to process a value, which means that multiple substrings can be masked independently by different ValueMaskers

Fixes #690

### DIFF
--- a/README.md
+++ b/README.md
@@ -1516,9 +1516,6 @@ for the path string format and more examples.  But in general:
 
 Specific values to be masked can be specified in several ways, as seen in the following example:
 
-When using regexes to identify strings to mask, all matches within each string field value will be replaced.
-If you want to match the full string field value, then use the beginning of line (`^`) and end of line (`$`) markers.
-
 ```xml
 <encoder class="net.logstash.logback.encoder.LogstashEncoder">
     <jsonGeneratorDecorator class="net.logstash.logback.mask.MaskingJsonGeneratorDecorator">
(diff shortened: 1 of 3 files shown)


brenuart (approved):
Looks ok to me.
Maybe add a note in the README + changelog about actual behaviour when multiple masks are configured.

PR #772, opened 2022-03-09
Make "%property{key}" return an empty string instead of "null" when used in PatternJsonProvider

### LINKED ISSUE #765: PatternJsonProvider: omitEmptyFields doesn't work with %property(ENV_VAR)
### Discussed in https://github.com/logfellow/logstash-logback-encoder/discussions/760

<div type='discussions-op-text'>

<sup>Excerpt from original post by **ziomill** February 16, 2022</sup>
I'm trying to use the "**omitEmptyFields**" to print/omit an ENV variable, without success. If the ENV_VAR 'CONTAINER_NAMESPACE' is not populated, the logger prints "null" as the field's value rather than omit the field. 
...
This is my configuration:

```xml
<encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
    <providers>
        ...
        <pattern>
             <omitEmptyFields>true</omitEmptyFields>
             <pattern>
                 {"container-namespace":"%property{CONTAINER_NAMESPACE}"}
             </pattern>
        </pattern>
        ...
    </providers>
</encoder>
```

And this is the printed log record:

```js
{
   ...
   "container-namespace":"null"
   ...
}
```
</div>

### Root Cause

The Pattern json provider is using a standard Logback PatternLayout under the hood to resolve placeholders and converters in the pattern string. The resulting string is then converted into JSON and added to the final result. The "omitEmptyFields" property works by telling Jackson to omit fields whose value is considered empty.

When a property does not exist, the `%property{}` converter returns `null` and this value is unfortunately added "asis" to the internal StringBuilder. In this example, the original pattern is converted into the string `{"container-namespace":"null"}` before it is transformed into a valid JSON structure by Jackson. Since the property is not null but is assigned a String value, it is not filtered out by Jackson's "empty field" filtering mechanism.

brenuart:
The behaviour of the `%property{key}` conversion specifier is slightly modified to return an empty string when the property is not defined instead of `null` as it normally does. Assuming the "foo" property does not exist, the following configuration:

```xml
<pattern>
   { "foo": "%property{foo}" }
</pattern>
```

... will produce something like this:
```json
{ "foo": "" }
```

... instead of:
```json
{ "foo": "null" }
```

The `foo` property will now be successfully omitted when `omitEmptyField` is `true`.

The property conversion specifier now also accepts an optional default value to use when the property does not exist. Assuming the property "foo" is not defined, `%property{foo:-bar}` will return `bar` instead of an empty string.

People wanting to revert to the original behaviour of returning "null" instead of an empty string can explicitly defined a default value of `"null"`.

This should close #765

### DIFF
--- a/src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
+++ b/src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
@@ -25,9 +25,9 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.classic.spi.LoggerContextVO;
 
 /**
- * Variation of the legacy Logback {@link PropertyConverter} with the option to specify a
- * default value to use when the property does not exist instead of returning {@code null}
- * as does the original Logback implementation.
+ * Variation of the Logback {@link PropertyConverter} with the option to specify a default
+ * value to use when the property does not exist instead of returning {@code null} as does
+ * the original Logback implementation.
  * 
  * <p>The default value is optional and can be specified using the <code>:-</code> operator as
  * in Bash shell. For example, assuming the property "foo" is not defined, <code>%property{foo:-bar}</code>
--- a/src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
+++ b/src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2013-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
```diff
@@ -0,0 +1,94 @@
...
+ * searched in the System environment.
+ * 
+ * 
+ * @author brenuart
+ */
+public class EnhancedPropertyConverter extends ClassicConverter {
+    /**
+     * Regex pattern used to extract the optional default value from the key name.
+     */
+    private static final Pattern PATTERN = Pattern.compile("(.+):-(.*)");
```
philsttr, line 49: might want to use `(.+?):-(.*)` here, to ensure the property name is before the first `:-`, in case `:-` appears later in the value.

src/main/java/net/logstash/logback/pattern/EnhancedPropertyConverter.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import ch.qos.logback.classic.pattern.ClassicConverter;
+import ch.qos.logback.classic.pattern.PropertyConverter;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.LoggerContextVO;
+
+/**
+ * Variation of the legacy Logback {@link PropertyConverter} with the option to specify a
```
philsttr, line 28: `legacy` seems a bit harsh, considering it is still used by logback.  I'd recommend removing the term legacy here

philsttr (commented):
Looks good!  Just a couple super minor things.

Issue #779, opened 2022-03-29
Provide Annotation Based Field Masker


frank-kiwi:
The FieldNameBasedFieldMasker is great to have and makes masking fields really easy.  Unfortunately, it is really difficult to configure and maintain for a big object model.

It would be great to be able to annotate the fields in the object model and have them automatically masked in the output.

philsttr:
See https://github.com/logfellow/logstash-logback-encoder/pull/782#issuecomment-1086726551

PR #782, opened 2022-03-30
#779 Add AnnotatedFieldMasker

frank-kiwi:
A new FieldMasker has been added which will allow the object model to be annotated and fields will be masked accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -1512,8 +1512,25 @@ for the path string format and more examples.  But in general:
 * Use `~1` to escape `/` within a token
 * Use `~0` to escape `~` within a token
 
-### Identifying field values to mask by value
+### Identifying field values to mask by Annotation
 
+```xml
+<encoder class="net.logstash.logback.encoder.LogstashEncoder">
+    <jsonGeneratorDecorator class="net.logstash.logback.mask.MaskingJsonGeneratorDecorator">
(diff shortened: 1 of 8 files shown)


philsttr:
Thank you for taking the time to submit this contribution @frank-kiwi !

While annotation support is an excellent idea for extending the masking support in logstash-logback-encoder, I cannot accept the current design into logstash-logback-encoder.

At this time I do not want to introduce classpath scanning or compile-time indexing requirements into logstash-logback-encoder.  Both classpath scanning and/or compile-time indexing introduce significant complexity and documentation requirements that I would prefer to be kept outside of logstash-logback-encoder.

Also, the implementation matches field names _anywhere_ in the output.  It does not guarantee that the field being written is actually part of the object whose class was annotated.   If I were a user of this library, I would expect the field to only be masked if it was part of the object whose class was annotated.  I would not expect it to match the annotated field name anywhere in the output (i.e. part of other objects).



As an alternative, you could try [configuring logstash-logback-encoder](https://github.com/logfellow/logstash-logback-encoder#customizing-json-factory-and-generator) to use a Jackson Mixin class for a specific domain class.  Then you could decorates sensitive fields in the Mixin class with `@JsonIgnore` or a custom serializer via `@JsonSerialize`.  This would not require any changes to logstash-logback-encoder.

PR #789, opened 2022-04-11
Fix NullPointerException in the "after the last transition" case

dongockien92:
When the timezone belong to the "after the last transition" case (ex: "Asia/Bangkok"), we'll get the `NullPointerException` at `FastISOTimestampFormatter`


brenuart:
@dongockien92 Thanks for this report and your proposed PR.
I discovered other edge cases when reviewing your solution and decided to create another branch whose content is heavily inspired from your PR.
I'm closing your PR now without merging it - please follow #794 for more information.

PR #806, opened 2022-05-09
Fixes memory leak on JDK 16+

### LINKED ISSUE #722: Probably memory leak in 7.0.1 version
**Describe the bug**
Some of our applications have been updated to the latest version `7.0.1` from `6.6` version. Later we found out one of the applications have a memory leak. We used heap dump to analyze it with Eclipse Memory Analyzer (https://www.eclipse.org/mat/) and found out that potentially this memory leak was introduced in the latest version of `logstash-logback-encoder`,

* logstash-logback-encoder version: 7.0.1
* logback version: 1.2.3
* jackson version: 2.12.5
* java version: 17.0.1

This is our logstash logback configuration:

```xml
    <property name="LOG_FILE" value="${LOG_FILE:-${LOG_DIR}/spring.log}"/>

    <appender name="FILE"
              class="ch.qos.logback.core.rolling.RollingFileAppender">
            <timeZone>UTC</timeZone>
            <throwableConverter class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
                <maxDepthPerThrowable>30</maxDepthPerThrowable>
                <exclude>sun\.reflect\..*\.invoke.*</exclude>
                <exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
                <rootCauseFirst>true</rootCauseFirst>
                <inlineHash>true</inlineHash>
            </throwableConverter>
        </encoder>
        <file>${LOG_FILE}</file>
        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
            <cleanHistoryOnStart>${LOG_FILE_CLEAN_HISTORY_ON_START:-false}</cleanHistoryOnStart>
            <fileNamePattern>${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
            <maxFileSize>${LOG_FILE_MAX_SIZE:-10MB}</maxFileSize>
            <maxHistory>${LOG_FILE_MAX_HISTORY:-7}</maxHistory>
            <totalSizeCap>${LOG_FILE_TOTAL_SIZE_CAP:-0}</totalSizeCap>
        </rollingPolicy>
    </appender>
```

And here are some memory leak reports:

Screenshot 2021-12-20 at 22 10 51

Screenshot 2021-12-20 at 22 11 27

Please let me know if I need to provide any additional information.

caesar-ralf:
In newer versions of java, specifically 16 and over, the `ThreadLocal` can
be cleared before the Thread is killed. This breaks the assumption of
`ThreadLocalHolder` which expected the `initializeThread` method to be
executed only once per `Thread`.

This commit changes the `threadValues` to become a map that holds which
thread the value is associated with so we can avoid recreating new
values when one already exists.

Fixes #722


caesar-ralf:
Ping @brenuart, could you take a look at this fix for the memory leak? I am sorry to ping you directly, but it's been a week and we are blocked off because of this problem.

src/main/java/net/logstash/logback/util/ThreadLocalHolder.java
```diff
@@ -148,39 +149,42 @@ public void close() {
...
+     *
      * @return a {@link Holder} assigned to the current thread.
      */
     private Holder<T> initializeThread() {
-        Holder<T> holder = new Holder<>();
-        threadValues.add(new HolderRef(Thread.currentThread(), holder, deadThreads));
-        return holder;
+        final Thread currentThread = Thread.currentThread();
+        final long threadId = currentThread.getId();
+        return threadValues.computeIfAbsent(
```
brenuart, line 181: I would add a comment here about the reason why the value associated with the current thread is stored in the threadValues Map and not in the ThreadLocal itself as it was before. Making reference to the ForkJoinPool issue clearing the thread local while the thread is still alive would be nice as well.

Note that this strategy makes the assumption that Thread.getId() does not change after a thread is created. This seems to be confirmed by the javadoc: the id seems to be immutable and unique...

brenuart (commented):
Looks nice. Relying on the Thread#getId and a Map to store the values instead of the ThreadLocal seems to be the only option.

One small general remark: your PR also contains cosmetic changes like removal of blanks... these changes produce some noise and make it harder to spot what is important :(

Thanks for your work. I'll merge it asap once you reviewed my other comments.

src/main/java/net/logstash/logback/util/ThreadLocalHolder.java
brenuart: Also, you need to rebase to pickup the latest changes otherwise the build will keep failing... (not related to your PR)

brenuart:
Also, you need to rebase to pickup the latest changes otherwise the build will keep failing... (not related to your PR)

caesar-ralf:
> One small general remark: your PR also contains cosmetic changes like removal of blanks... these changes produce some noise and make it harder to spot what is important :(

ow yeah, sorry for that. That was my IDE doing automatically. I will work on fixing the issues raised by the PR

src/main/java/net/logstash/logback/util/ThreadLocalHolder.java
caesar-ralf: added

caesar-ralf:
> Also, you need to rebase to pickup the latest changes otherwise the build will keep failing... (not related to your PR)

Changed it. I checked why it was failing in jdk8 and it seemed to be a hiccup because it couldn't download the javadocs. Feel free to change the comment I added as you please

brenuart:
Merged. Thanks for your work.

caesar-ralf:
Thanks for helping out @brenuart and being understanding! Can you let me know when you do release with this fix?

brenuart:
Soon (hopefully by the end of the week). Just waiting for another PR to be merged and we are all set ;-)

fredericdasilva:
Thanks a lot. This commit fixed memory leaks that we faced in Production since last upgrade on JDK 17.

Issue #829, opened 2022-08-09
Add support for Timestamp with nanosecond precision (logback 1.3.0)


brenuart:
Starting from 1.3.0-alpha12, Logback events are now storing timestamp with nanosecond precision (using an Instant). The `AbstractTimestampJsonProvider` should be updated to take this enhancement into account.

**(1) Backward compatibility**

A new `getInstant()` method has been added to `LoggingEvent` in version 1.3.0 which was not available in previous versions. The implementation should make use of reflection to detect at runtime if the `getInstant()` method is available and automatically enable nanosecond precision in our JsonProvider.

It should be noted that as of version 1.3.0-beta0 AccessEvents have not (yet) been enhanced with a `getInstant()` method and therefore lack nanosecond precision. Using the above mentioned strategy should help us to keep using the current `AbstractFormattedTimestampJsonProvider` as a base class for both the LoggingEvent and AccessEvent specialisations: `getInstant()` not available on AccessEvent -> nanosecond not enabled -> revert to millis as currently.

**(2) Available timestamp formats and default**

By default timestamps are formatted as [ISO_OFFSET_DATE_TIME]. As per the javadoc of `java.time.format.DateTimeFormatter` this format outputs nanoseconds but drops the insignificant leading zeros. This means that:
- precision will be at most three digits (milliseconds) when used with logback 1.2
- precision will be up to 9 digits (nanoseconds) when used with logback 1.3
Users should be aware they will suddenly have a higher timestamp precision just by upgrading the logback and LLE versions while keeping the same configuration. If nanosecond precision is not desired, users have the option of configuring an explicit format pattern instead of using the [ISO_OFFSET_DATE_TIME] keyword.

`[UNIX_TIMESTAMP_AS_NUMBER]` and `[UNIX_TIMESTAMP_AS_STRING]` will behave asis.

brenuart:
@philsttr What's your opinion? Any feedback?

PR #832, opened 2022-08-14
Bump to Logback 1.3.0-beta0

brenuart:
- Build against Logback 1.3.0
- Execute backward compatibility tests against latest Logback 1.2.x
- Comment links to Logback javadoc (generation fails under Java8 because logback docs have replace "package-list" by "element-list")


brenuart:
closes #829

PR #868, opened 2022-10-04
allow set a flag to compile the pattern

ahernandez-technisys:
This PR allows to set a flag when configuring value masker ex:

```xml
<valueMask>
    <value>(?i)foo</value>
    <mask>*********</mask>
    <flag>8</flag>
</valueMask>
```

this example compiles the pattern with multiline


brenuart:
Hi @ahernandez-technisys,

Thanks for this contribution.
As you noticed, it is already possible to set/unset the flags directly in the regex pattern (feature supported by the Java Pattern itself). Why do you need another way of specifying flags?

PR #968, opened 2023-06-07
Honor overriden `writeObject` of `JsonGenerator` passed to `setJsonGeneratorDecorator`

rdesgroppes:
The order in which `JsonGeneratorDelegate`s are chained currently leads to disregard the `writeObject` method of the `JsonGenerator` passed to `setJsonGeneratorDecorator`.

The internal `SimpleObjectJsonGeneratorDelegate` indeed intercepts `writeObject` for quite a number of concrete types:
https://github.com/logfellow/logstash-logback-encoder/blob/933a0f51c1bfefc838adee9982ad3537a08a39bd/src/main/java/net/logstash/logback/util/SimpleObjectJsonGeneratorDelegate.java#L38-L60
```java
// etc.
```

The present change therefore consists in moving any user-defined `JsonGenerator` to the top level of the delegation chain, leaving users the power (and therefore the responsibility) to adjust the output to their needs.


rdesgroppes:
Hi @brenuart, may I kindly request a review? Merci !

PR #970, opened 2023-06-17
Drop support for logback 1.2

philsttr:
Drop support for logback 1.2 in preparation for upcoming support for keyValue pairs in slf4j 2 and logback 1.3


analytically:
This is quite some breaking change.

carl-parrish:
upgrading from 7.3 to 7.4 just destroyed all my unitTests so far this is the only thing I can think that would have that kind of impact. I didn't write the original code any suggestions on what I should look for to see if I had logbook 1.2 dependencies in my code? Any docs on migration?

philsttr:
Hi @carl-parrish, Check the [logback news](https://logback.qos.ch/news.html) for details on changes in logback to 1.3.

PR #972, opened 2023-06-18
Exclude configured MDC keys from processing with MdcEntryWriters

jug:
@philsttr You were a little too fast with merging & releasing ;-)

Follow up for #957 providing a way to skip processing of the MdcEntryWriters.

The use for this comes into play as we have lots of services with a standard prefixed set of MDC keys for which for example the automatic MDC long-value writing should not happen.

And as a configuration is preferred over custom implementations, here's a proposal to configure this (with an exclude pattern).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1063,6 +1063,16 @@ Currently, MDC entry writers for the following value types are supported:
 To add your own MDC entry writer for other types or apply the manipulations only for specific fields
 you can write your own implementation of [`MdcEntryWriter`](src/main/java/net/logstash/logback/composite/loggingevent/mdc/MdcEntryWriter.java).
 
+Writing MDC entries with the `MdcEntryWriters` can be skipped for MDC keys matching an exclude pattern
+(which is a regular expression with the syntax of `java.util.regex.Pattern`):
+
+```xml
+<encoder class="net.logstash.logback.encoder.LogstashEncoder">
+    <mdcEntryWriterExcludeKeyPattern>(skipExactKey|skipPrefix\\..*)</mdcEntryWriterExcludeKeyPattern>
+</encoder>
+```
+By default, all MDC entries are written as String values.
+
 You can also replace the default MDC JSON provider with your own class extending from
 [`MdcJsonProvider`](src/main/java/net/logstash/logback/composite/loggingevent/MdcJsonProvider.java).
 Configuring your class as a [Custom JSON Provider](#custom-json-provider) will then replace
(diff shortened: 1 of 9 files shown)


philsttr:
This feature could be easily confused with the existing `includeMdcKeys` / `excludeMdcKeys` already on the `MdcJsonProvider`.

Perhaps it would be more obvious if it was implemented as a composite `MdcEntryWriter` that delegates to other `MdcEntryWriter`s if the entry meets certain include/exclude criteria.  something like

```java
public class RegexFilteringEntryWriter implements MdcEntryWriter {

    private final List<Pattern> includeMdcKeyPatterns = new ArrayList<>();
    private final List<Pattern> excludeMdcKeyPatterns = new ArrayList<>();
    private final List<MdcEntryWriter> mdcEntryWriters = new ArrayList<>();

    @Override
    public boolean writeMdcEntry(JsonGenerator generator, String fieldName, String mdcKey, String mdcValue) throws IOException {
        
        if (matches includes and doesn not match excludes) {
            for (MdcEntryWriter mdcEntryWriter : this.mdcEntryWriters) {
                if (mdcEntryWriter.writeMdcEntry(generator, fieldName, mdcKey, mdcValue)) {
                    return true;
                }
            }
        }
        return false;
    }

    // ... snip ...
}
```

jug:
@philsttr 
philsttr> This feature could be easily confused with the existing `includeMdcKeys` / `excludeMdcKeys` already on the `MdcJsonProvider`.

Before this pull-request I had tried another approach of having a base-class for the `MdcEntryWriter` with a common configuration: https://github.com/jug/logstash-logback-encoder/commit/c16b96bdcfcfe65ff5f324cfe49d61356dccb3e4

But I was not happy with it, because it would require to copy excludes into multiple MDC writers (using `LongMdcEntryWriter` together with `DoubleMdcEntryWriter` you need to exclude the same key from both writers as a long value would also be parsed as double).

But yikes, I think I had no imagination how flexible the `MdcEntryWriter` could really be with your proposal for a nested approach...

philsttr> Perhaps it would be more obvious if it was implemented as a composite `MdcEntryWriter` that delegates to other `MdcEntryWriter`s if the entry meets certain include/exclude criteria. something like
> 
> ```java
> public class RegexFilteringMdcEntryWriter implements MdcEntryWriter {
> 
>     private final List<Pattern> includeMdcKeyPatterns = new ArrayList<>();
>     private final List<Pattern> excludeMdcKeyPatterns = new ArrayList<>();
>[...]

philsttr> This would move the properties to the writer, so it's obvious that they apply only to the contained writers.
philsttr> It would also allow configuring multiple `RegexFilteringMdcEntryWriter` instances [...]

I like this, as it would also solve the issue in the other branch (mentioned above).

Though I'm not sure about the `List` of `Pattern`. Just one `Pattern` for include and exclude should suffice as multiple patterns can be merged into just one regex.

jug:
@philsttr Opened new pull request with proposal for `RegexFilteringMdcEntryWriter` in #974

Feel free to review.

philsttr:
Closing in favor of #974

PR #988, opened 2023-09-12
Set AbstractThrowableClassNameJsonProvider.getThrowable as protected

EvaristeGalois11:
Hi, I'm trying to customize the `throwable_class` field name but I found out that it isn't possible to extend the class `AbstractThrowableClassNameJsonProvider` because its abstract method `getThrowable` is package protected.

I think it doesn't make much sense to have a public abstract class that isn't really publicly extendable. Also this should align the class with the similar `AbstractThrowableMessageJsonProvider`, which already has its abstract method as protected.

As a temporary workaround I'm extending the class `ThrowableClassNameJsonProvider`, this way the package protected method is already implemented and the problem is circumvented.

Let me know what you think about this!

Bye!


philsttr:
Looks good.  Thanks for the contribution!

PR #1043, opened 2024-08-18
Make it possible to pretty print throwables as array of strings.

### LINKED ISSUE #1042: Prettier viewing of stacktrace
**Is your feature request related to a problem? Please describe.**
Make it easier to view exceptions when pretty print option is on.

**Describe the solution you'd like**
Instead of outputting stacktrace as a single string, output as an array of strings where each string is a stacktrace line.

maxxedev:
fixes #1042


laurentthiebaudezm:
That will change my life thanks

PR #1086, opened 2025-10-09
Allow setting droppedWarnFrequency=0 to disable logging dropped message warnings

tommyulfsparre:
Currently there is no way to completely disable logging of dropped events. 

Using `droppedWarnFrequency` as suggested in the `README` isn't possible because setting the value to 0 will lead to a division by zero exception at [AsyncDisruptorAppender.java#L514](https://github.com/logfellow/logstash-logback-encoder/blob/main/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java#L514). 

>  When the appender drops an event, it emits a warning status message every droppedWarnFrequency consecutive dropped events (1000 by default, use 0 to turn off warnings). Another status message is emitted when the drop period is over and a first event is succesfully enqueued reporting the total number of events that were dropped.

Setting `droppedWarnFrequency` to something like `Integer.MAX_VALUE` doens't work either because the appender will emit how many total event that was dropped on the first successful enqueue after a ringbuffer full event.

This PR fixes it so that setting droppedWarnFrequency=0 will disable the log warnings, as mentioned in the readme

### DIFF
--- a/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
+++ b/src/main/java/net/logstash/logback/appender/AsyncDisruptorAppender.java
@@ -259,11 +259,6 @@ public abstract class AsyncDisruptorAppender<Event extends DeferredProcessingAwa
     private Duration shutdownGracePeriod = Duration.buildByMinutes(1);
 
     /**
-     * When false, no logging will occur for dropped events.
-     */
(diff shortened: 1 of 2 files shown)


tommyulfsparre:
@philsttr Much better, thank you!

Issue #1083, opened 2025-09-13
A terminally deprecated method in sun.misc.Unsafe has been called


pgehl:
**Describe the bug**
Upon service startup we see this warning message:

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::arrayBaseOffset has been called by net.logstash.logback.encoder.com.lmax.disruptor.RingBufferFields (file:/workspace/BOOT-INF/lib/logstash-logback-encoder-8.1.jar)
WARNING: Please consider reporting this to the maintainers of class net.logstash.logback.encoder.com.lmax.disruptor.RingBufferFields
WARNING: sun.misc.Unsafe::arrayBaseOffset will be removed in a future release
```

**To Reproduce**
1. Use this LogstashTcpSocketAppende configuration...
```
        <appender name="LOGSTASH" class="net.logstash.logback.appender.LogstashTcpSocketAppender">
            <destination>${LOGSTASH_DESTINATION}</destination>
            <encoder class="net.logstash.logback.encoder.LogstashEncoder">
            </encoder>

            <!-- logs a warning on the first success after a series of consecutive append/send/connect failures. -->
            <listener class="net.logstash.logback.appender.listener.FailureSummaryLoggingAppenderListener"/>
        </appender>

```

**Expected behavior**
No jvm unsafe warning when using the appender

**Additional context**
We run java spring-boot applications inside a kubernetes cluster

* logstash-logback-encoder version: 8.1
* logback version: 1.5
* jackson version: 1.5.18
* java version: 24

philsttr:
This was coming from the shaded version of lmax disruptor.   The issue was fixed by https://github.com/LMAX-Exchange/disruptor/pull/392 in lmax disruptor 4.0.0.

Will be fixed in logstash-logback-encoder by upgrading to lmax disruptor to 4.0.0 in #1096

philsttr:
Fixed in #1096

PR #1096, opened 2025-10-19
Upgrade lmax disruptor to 4.0.0

philsttr:
disruptor 4.0.0 folded LifecycleAware and SequenceReportingEventHandler into EventHandler interface


philsttr:
also fixes #1083

PR #1121, opened 2026-04-22
Jackson 3.1.2 update

NoZomIBK:
Updated from https://github.com/logfellow/logstash-logback-encoder/pull/1119 to 3.1.2, so javadoc is included again


philsttr:
Closing in favor of #1110.

In general, there's no need to create PRs for dependency-only updates, since this repo uses dependabot to manage them.

